Issue #19, opened 2022-04-06
TimdexQueryWrapper


JPrevost:
Success Criteria

- Takes a pre-built TIMDEX query as input
- Calls TIMDEX V2 API
- Handles Fatal Errors
- Returns Reponse

Out of scope

- Analyzing results
- Converting results into Facets, Records
- Pagination (this is handled in the QueryBuilder and UI components)

Issue #18, opened 2022-04-06
TimdexQueryBuilder


JPrevost:
Success Criteria

- Transforms simple Query to a call that can be sent to TIMDEX V2 API
- Pagination

Out of scope

- Advanced queries, including faceting, targeting specific collections, targeting specific fields

matt-bernhardt:
Ah shoot, the magic commit message didn't catch this one. It was done as part of #23

PR #23, opened 2022-04-06
First steps toward Query Wrapper

matt-bernhardt:

* We need a wrapper model for the Timdex API in order to easily send and
process searches against it.
* We also need a separate query builder which can support future
expansion.

### Relevant ticket(s):

* https://github.com/MITLibraries/timdex-ui/issues/18
* https://github.com/MITLibraries/timdex-ui/issues/19

#### TimdexWrapper

* This builds a TimdexWrapper model with a single public method, search.
There are private methods for reading the two environment variables,
as well as for extending the TIMDEX_BASE value to the specific search
endpoint (using the result of the query builder)
* I've chosen to use private methods for all of these as I don't think
the rest of the application will need to interact with those tools
separately? This may not end up being the case, in which case they can
be made public (and have tests written for them).
* The tests for the wrapper use VCR cassettes, and include a limited
range of search types. These can be expanded, but they do risk
duplicating the tests for the querybuilder. The tests should be robust
enough that they don't depend on specific result counts, which should
help needing to constantly update them.
* There is some limited error handling, but more will be needed before
we launch. Currently there is only one rescue step for all errors,
which returns the error key and message.
* The tested errors are Timdex being down, and a returned 404 response.

#### QueryBuilder

* The QueryBuilder model currenty takes a string input, returning a
Hash for all parameters that get passed to Timdex. Future work will
expand this significantly, to handle advanced search and other use
cases. For now, though, this is pretty stripped down.
* The tests for the query builder deal with a similarly limited set of
search cases (multiple words, quotes, and trimming). Eventually
these tests will (should) diverge from what the timdex wrapper is
testing for, but for now they overlap a lot.

* Hopefully none by this point

### Developer

### Code Reviewer

NO

YES

* annotate
* dotenv-rails
* http
* vcr
* webmock

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,3 +3,8 @@
 # TIMDEX UI
 
 A discovery interface backed by [the TIMDEX API](https://github.com/MITLibraries/timdex).
+
+## Optional Environment Variables
+
+- `TIMDEX_BASE`: value to override the default url for the TIMDEX API.
+- `TIMDEX_TIMEOUT`: value to override the 6 second default for TIMDEX timeout.
--- a/app/models/timdex_wrapper.rb
+++ b/app/models/timdex_wrapper.rb
@@ -1,0 +1,44 @@
+class TimdexWrapper
+  def initialize
+    @timdex_http = HTTP.headers(accept: 'application/json',
+                                'Accept-Encoding': 'gzip, deflate, br',
+                                'Content-Type': 'application/json',
+                                Origin: 'https://lib.mit.edu')
+  end
+
+  # Run a search
+  # @param term [string] The string we are searching for
+  # @return [Hash] A Hash with search metadata and an Array of {Result}s
+  def search(term)
+    @results = @timdex_http.timeout(http_timeout)
+                           .get(search_url(term))
+    JSON.parse(@results.to_s)
+  rescue StandardError => e
+    error = {}
+    error['error'] = e.to_s
+    error
+  end
+
+  private
+
+  # https://github.com/httprb/http/wiki/Timeouts
+  def http_timeout
+    if ENV['TIMDEX_TIMEOUT'].present?
+      ENV['TIMDEX_TIMEOUT'].to_f
+    else
+      6
+    end
+  end
+
+  def search_url(term)
+    timdex_url + "search?#{QueryBuilder.new(term).querystring}"
+  end
+
+  def timdex_url
+    if ENV['TIMDEX_BASE'].present?
+      ENV['TIMDEX_BASE'].to_s
+    else
+      'https://timdex.mit.edu/api/v1/'
+    end
+  end
+end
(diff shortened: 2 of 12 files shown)


app/models/timdex_wrapper.rb
```diff
@@ -0,0 +1,57 @@
+class TimdexWrapper
+  def initialize
+    @timdex_http = HTTP.persistent(timdex_url)
```
JPrevost, line 3: I'm not sure how long you intend the TimdexWrapper to stay around after initialization. If it's for a single query, I think we should remove the persistent connection.

If you have in mind a way to initialize once and re-use the same instance of TimdexWrapper across multiple searches the persistent connection is a value for sure.

If you aren't sure yet we can talk it through and decide which will make the most sense based on what we know so far.

app/models/timdex_wrapper.rb
```diff
@@ -0,0 +1,57 @@
...
+  # https://github.com/httprb/http/wiki/Timeouts
+  def http_timeout
+    if ENV['TIMDEX_TIMEOUT'].present?
+      ENV['TIMDEX_TIMEOUT'].to_f
+    else
+      6
+    end
+  end
+
+  def query(term)
```
JPrevost, line 38: We had split out this functionality into a TimdexQueryBuilder class. I know that seems silly based on the current needs, but since we know that the QueryBuilder will eventually get quite complex I'd prefer to add the abstraction early and build on it rather than have to move this later (which we may keep putting off until we have a mega "search" model.

README.md
```diff
@@ -3,3 +3,8 @@
 # TIMDEX UI
 
 A discovery interface backed by [the TIMDEX API](https://github.com/MITLibraries/timdex).
+
+## Optional Environment Variables
+
+- `TIMDEX_V1`: value to override the default url for the TIMDEX API.
```
JPrevost, line 9: I'd like us to consider using something like `TIMDEX_BASE_URL` or something instead of `TIMDEX_V1` as it'll get confusing to use set the v2 base in that ENV :)

JPrevost (changes requested):

README.md
matt-bernhardt: Yeah, I wasn't sure if there was going to be a need for us to have both APIs in env simultaneously. If we don't think that's the case, then yeah something generic like `TIMDEX_BASE_URL` makes more sense.

JPrevost: That's a good point. I'm not sure we've decided if this UI will be v1/v2. I think we probably want to just have it be v2 or we triple our work and just delete it all in a few months.

app/models/timdex_wrapper.rb
matt-bernhardt: Honestly, I grabbed this approach from Bento but didn't think critically about it. Looking over documentation now.

Issue #20, opened 2022-04-06
Enhancer


JPrevost:
Success Criteria

- Initial model for Enhancer

Out of scope

- Actually checking for any patterns

PR #27, opened 2022-04-13
Adds initial Enhancer class and two examples

JPrevost:

* We want an way to detect intent from user input and believe analyzing
  the raw input in various ways and categorizing what we know about it
  will be helpful in both building up queries to our search indexes but
  also in detecting when something other than a search index may be
  helpful

Relevant ticket(s):

* https://github.com/MITLibraries/timdex-ui/issues/20

* Creates an Enhancer class and EnhancerPatterns class with two initial
  examples (ISBN and ISSN detection)

- this is not yet connected to the user input stage or query builder
  stage as there is work underway already that this will somewhat
  conflict with so we'll need to wire it up when it up separately

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/models/enhancer.rb
+++ b/app/models/enhancer.rb
@@ -1,0 +1,16 @@
+class Enhancer
+  attr_accessor :enhanced_query
+
+  # accepts all params as each enhancer may require different data
+  def initialize(params)
+    @enhanced_query = {}
+    @enhanced_query[:q] = params[:q] if params[:q]
+    patterns(params) if params[:q]
+  end
+
+  private
+
+  def patterns(params)
+    @enhanced_query = EnhancerPatterns.new(@enhanced_query, params[:q]).enhanced_query
+  end
+end
--- a/app/models/enhancer_patterns.rb
+++ b/app/models/enhancer_patterns.rb
@@ -1,0 +1,28 @@
+class EnhancerPatterns
+  attr_accessor :enhanced_query
+
+  def initialize(enhanced_query, term)
+    @enhanced_query = enhanced_query
+    term_pattern_checker(term)
+  end
+
+  private
+
+  def term_pattern_checker(term)
+    term_patterns.each_pair do |type, pattern|
+      @enhanced_query[type.to_sym] = match(pattern, term) if match(pattern, term).present?
+    end
+  end
+
+  def match(pattern, term)
+    pattern.match(term).to_s.strip
+  end
+
+  # term_patterns are regex patterns to be applied to the basic search box input
+  def term_patterns
+    {
+      isbn: /(ISBN-*(1[03])* *(: ){0,1})*(([0-9Xx][- ]*){13}|([0-9Xx][- ]*){10})/,
+      issn: /[0-9]{4}-[0-9]{3}[0-9xX]/
+    }
+  end
+end
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@
 
 # dotenv
 .env
+.env.development
 
 ## Environment normalization:
 /.bundle
@@ -33,3 +34,4 @@ yarn-debug.log*
 /config/master.key
 
 .DS_Store
+.vscode/
--- a/test/models/enhancer_patterns_test.rb
+++ b/test/models/enhancer_patterns_test.rb
@@ -1,0 +1,77 @@
+require 'test_helper'
+
+class EnhancerPatternsTest < ActiveSupport::TestCase
+  test 'ISBN detected in a string' do
+    enhanced_query = {}
+
+    actual = EnhancerPatterns.new(enhanced_query, 'test 978-3-16-148410-0 test').enhanced_query
+    assert_equal('978-3-16-148410-0', actual[:isbn])
+  end
+
+  test 'ISBN-10 examples' do
+    enhanced_query = {}
+
+    # from wikipedia
+    samples = ['99921-58-10-7', '9971-5-0210-0', '960-425-059-0', '80-902734-1-6', '85-359-0277-5',
+               '1-84356-028-3', '0-684-84328-5', '0-8044-2957-X', '0-85131-041-9', '93-86954-21-4', '0-943396-04-2',
+               '0-9752298-0-X']
+
+    samples.each do |isbn|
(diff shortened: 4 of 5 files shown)


app/models/enhancer.rb
```diff
@@ -0,0 +1,17 @@
+class Enhancer
+  attr_accessor :enhanced_query
+
+  # accepts all params as each enhancer may require different data
+  def initialize(params)
+    @enhanced_query = {}
+    @enhanced_query[:raw_query] = params
```
matt-bernhardt, line 7: Some part of me is nervous about blindly trusting everything that a user might pass in as a querystring - would it be preferrable to deal only with parameters which we know and expect might exist here?

app/models/enhancer.rb
```diff
@@ -0,0 +1,17 @@
...
+    @enhanced_query[:q] = params[:q] if params[:q]
+    patterns(params) if params[:q]
+  end
```
matt-bernhardt, line 10: If I understand what you're setting up here, enhancers which are different than regex matches would get a different method here, and a different subclass? This would then anticipate things like the hints system, or journal name matching, or timely-news detection?

app/models/enhancer_patterns.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  def match(pattern, term)
+    pattern.match(term).to_s.strip
+  end
+
+  # term_patterns are regex patterns to be applied to the basic search box input
+  def term_patterns
+    {
+      isbn: /(ISBN-*(1[03])* *(: ){0,1})*(([0-9Xx][- ]*){13}|([0-9Xx][- ]*){10})/,
```
matt-bernhardt, line 24: I'm not familiar with isbn and issn regex at all, but I do believe that they _can_ be detected by regex. Provided we got these from their spec or something authoritative, then this is great. I like this sort of extendable data structure.

matt-bernhardt (commented):
None of these comments are in the vein of "please make this change" - although I'm most concerned about the first comment about trusting params too much. The other two comments are just questions to make sure I understand our intent, and to confirm how I think we're getting the regex patterns.

app/models/enhancer.rb
JPrevost: Yup

app/models/enhancer_patterns.rb
JPrevost: Yeah both of these patterns are simple enough for this initial test. The issn pattern comes from Wikipedia (and is super easy). The isbn pattern from some a regex site (and is somewhat more complex but mostly only if we care about things we generally won't).

Of note, both specs include checksums to ensure they are valid, but for our use case of "does this look like an issn/isbn" I don't think that's useful. The patterns themselves (which wikipedia talks about extensively) should be good enough for us to decide if we want to run a modified search or not.

app/models/enhancer.rb
JPrevost: This is just a convenient way to leave the unmodified version around. We will still just pull out what we expect to be used in various analyzers/builders. Does that help or am I missing something about what you are concerned with?

app/models/enhancer_patterns.rb
matt-bernhardt: If we end up wanting to include the checksum portion, that feels like something we can add later - so yeah I'm not worried about that issue at the moment.

There may be other needs to tweak this depending on its implementation in the near- and medium- future, but for now I think this is fine based on what you're saying. Thanks!

app/models/enhancer.rb
matt-bernhardt: I'm worried about future-us doing something with untrusted user input, and whether it being an attribute of an `enhanced_query` might lull us into thinking that it is trustworthy. The "raw" prefix in the name should help this, I think - and I may be overthinking this in general.

If I understand the application flow correctly, the intent is that the enhanced_query hash is what's going to be sent to the QueryBuilder, and then to the TimdexWrapper? Somewhere along the line, I think I'd prefer for there to be some sort of whitelist/filter for fields - particularly if we're going to eventually pass user input down into the Timdex UI or other deeper systems, especially of their APIs aren't read-only.

(Although I'm also realizing that Timdex is a read-only system, so I don't think it matters what we throw at it via a GET request?)

I may be overthinking this - if all of this doesn't resonate with you, then I'm happy to approve the PR as-is.

JPrevost: Not trusting user input as a default stance is good security practice. However, copying the params to another variable and including it in a hash changes nothing to our security practices. If renaming to make it clear that `raw_query` is indeed just the request params, I'd be happy to do that. Would `raw_params` or `request_params` feel less likely to be mistaken for safe data to you?

For additional context, I felt we had two options:

- pass the params around separately from the enhanced_query as things downstream will need different bits and we don't know what yet
- store the params inside the enhanced_query so we only pass one thing around and downstream methods that need to pull a param can do so but things that don't won't need to even know params exist

I'm also not opposed to moving towards pulling out everything we need like we do `q` (essentially creating the safelist of params but not actually without explicit action making sure they are actually safe) but I think I'd like to sit on that for a bit as that seems more likely to be a concern of the QueryBuilder and not the Enhancers (although the Enhancers might end up caring about additional params... but I'm not sure what yet?)

Finally, nothing I am doing in this PR uses `raw_query`. I don't mind removing it entirely until we get a better feel for whether it makes sense. It was added as "I suspect this will be useful later" but it would be super easy to add back in later if that's the case.

matt-bernhardt: I'd prefer the name `raw_query` over `request_params`, which you're already using - and I'm not coming up with a better alternative that doesn't introduce its own problems.

While it might be preferable to introduce this only when we have a way of using it, we're at the stage of the project where things aren't expected to be super clean and tidy? If you think there's going to be a general cleanup in the future to optimize things, then I'd say we can merge this now.

In that vein, I'll approve the PR as-is.

matt-bernhardt (approved):
After discussion (thanks!) I think this is good to go.

:shipit:

app/models/enhancer.rb
JPrevost: Thanks for this conversation. For now I'm going to remove the unused affordance and when/if we need it we'll have more context for whether it introduces concerns and what to do about them.

PR #33, opened 2022-04-22
Displays core datacite fields

JPrevost:

* The data ingest and api work has started with these fields so UI is doing the same

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/RDI-94

* Display the core datacite fields if they exist for records

* Cassettes now strip out the timdex URL during generation to allow more easily using local, staging, or production timdex to generate necessary cassettes while making them look all the same so replays are consistent
* Hardcoded "origin" has been changed to a variable. I'm not sure what we are even usign this for but instead of removing it I allowed it to be configured.
* All tests now use v2 cassettes which required some minor changes to tests and regeneration of all cassettes
* One test is skipped that wasn't using a cassette but looks like it thinks it was
* The UI is still janky. We'll want to add intentional structured HTML and CSS soon.

#### Developer

#### Code Reviewer

NO

YES


jazairi (approved):
Looks good. :shipit: I have a couple questions about the result partial that I think would be best discussed in RDI-82.

PR #36, opened 2022-05-11
Apply initial theming

matt-bernhardt:

* The theming principles identified already as part of RDI-82 need to be
applied to the application.

#### Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/rdi-97

This looks like a big diff, but the number of affected files is
deceptive. The changes break down as follows:

* Adding the Roboto font in order to locally store files and save users
from having to interact with Google Fonts ( app/assets/fonts/ )
* Adding new CSS partials to tweak rules from the theme gem according to
RDI-82. Eventually the rules in timdexui.scss will hopefully find new
homes...
* Adding a static page to support a local style guide
( app/views/static , the static controller, and routes.rb )
* Updates to the basic search views:
  * Changing result list back to a list markup with headings
  * Updates to page heading structure (results are now h2s, not h4s -
and facet categories change from h4 to h3)
  * Applying new styles to the fact panels, search form, and fonts
  * Search form is slighty responsive now depending on whether the user
has already done a search or not.
  * Updates to basic search controller tests because some markup is
changing.

Because this ticket is only focused on overall theming changes, there
are still some parts of the stylesheet that have yet to change - buttons
are one prominent example. The hope is that deeper consideration of the
search results (RDI-99) and other UI details (RDI-100) will provide the
space for this type of work.

* I don't think there are any side effects. No models change, and the
only controller change is the addition of the static controller.

#### Developer

#### Code Reviewer

NO

NO


app/assets/stylesheets/fonts.scss
```diff
@@ -0,0 +1,27 @@
+/* roboto-regular - latin */
+@font-face {
+  font-family: 'Roboto';
+  font-style: normal;
+  font-weight: 400;
+  src: url('roboto-v29-latin-regular.eot'); /* IE9 Compat Modes */
+  src: local(''),
+       url('roboto-v29-latin-regular.eot?#iefix') format('embedded-opentype'), /* IE6-IE8 */
```
jazairi, line 8: Kinda curious as to whether we need shims that go back this far. (Of course, now that I've said it, we'll see thousands of IE6 user agents in the logs.)

app/assets/stylesheets/application.css.scss
```diff
@@ -3,5 +3,11 @@
...
 
 @import "libraries-main";
+@import "fonts";
+@import "partials/_variables";
+@import "partials/_alerts";
 @import "partials/_facets";
+@import "partials/_panels";
 @import "partials/_search";
+@import "partials/_typography";
+@import "timdexui";
```
jazairi, line 13: I checked our other rails apps out of curiosity, and I noticed that in Bento the 'main' stylesheet is also a partial, whereas in Thing there are no partials. Not recommending a change here, since this approach makes logical sense to me, but I wonder if that reaches the level of needing a team discussion about consistency in how we load stylesheets.

jazairi (approved):
What a step forward in our theming! Great work on this. I'm looking forward to seeing what our colleagues in UX think; hopefully we can move some of these changes upstream in time.

Thanks as always for the detailed commit message, which made review much easier.

:shipit:

matt-bernhardt:
Thanks! I agree with having a team discussion on topics like how we load stylesheets - I'll put that (and the font support issue) in our backlog.

PR #38, opened 2022-05-17
Load results from GraphQL rather than REST endpoint

matt-bernhardt:

* We have decided to use the GraphQL API for this UI, rather than the
  REST API.

This converts the search path of the application to load results from
the GraphQL endpoint. This involves a number of specific changes:

* Defines a new Timdex model that wraps the GraphQL endpoint. Eventually
  the TimdexWrapper model will go away entirely, but at the moment it is
  still used for the full record route.
  * The new Timdex model uses the equivalent of parameterized queries to
    better protect against GraphQL injection.
* Updates the basic_search views to use the GraphQL field names (which
  are camelCased rather than snake_cased)
* Adds an error template for cases when the API returns an error. This
  will need further refinement in future tickets however.
* Adds a cache of the GraphQL schema in config/schema/schema.json. A
  future ticket will be needed to define how this gets updated.
* Updates a number of cassettes to reflect this API change.
* Because the new wrapper handles user parameters differently (and IMO
  better), some cassettes are no longer needed because the tests they
  supported are no longer necessary.

* For the moment, at least, the Timdex model is built differently from
  other models in the application, making extensive use of Contants
  instead of methods. This follows the documentation for the graphql-
  client gem - which introduces a different way of working with this
  model compared to the rest of the application. This isn't ideal, but
  I've been unable to get a fully working implementation of this gem
  using a model with methods. I'll push up a separate commit on a
  different branch to demonstrate my progress downt his path, but it
  is not ready yet.

#### Ticket

https://mitlibraries.atlassian.net/browse/RDI-104

#### Developer

#### Code Reviewer

NO

YES


matt-bernhardt:
Please note - I've prepared a separate PR that extends this work towards a method-based class. The code works in the console, but something about integrating the class with the controller is continuing to fail.

You can see that work, which includes the console commands I've been attempting to use, at https://github.com/MITLibraries/timdex-ui/pull/38

app/models/timdex.rb
```diff
@@ -0,0 +1,70 @@
...
+require 'graphql/client/http'
+
+class Timdex
+  HTTP = GraphQL::Client::HTTP.new(ENV.fetch('TIMDEX_GRAPHQL', '')) do
+    def headers(*)
+      { 'User-Agent': 'MIT Libraries Client' }
+    end
+  end
+
+  Schema = GraphQL::Client.load_schema('config/schema/schema.json')
```
JPrevost, line 11: Can you document how to regenerate this file in the README and provide a bit of context as to when it needs to be regenerated (which I think is largely whenever we update the the GraphQL schema upstream)

JPrevost (changes requested):
See comment about added documentation around the timdex schema generation (example console commands to regenerate it would be great).

app/models/timdex.rb
matt-bernhardt: Oh yeah - that's a good suggestion. Second commit underway...

JPrevost (approved):

PR #39, opened 2022-05-20
Speculative implementation of method-based timdex wrapper

matt-bernhardt:
This creates a parallel model of the Timdex class, using methods
rather than constants where possible. The parallel, TimdexCandy,
will work in the console - but I've been unable to get it working
from the controller.

The steps in the console are:

$ query = {'q' => 'data'}
$ wrapper = TimdexCandy.new
$ Parsed = wrapper.client.parse(TimdexCandy::SearchQueryText)
$ wrapper.search(Parsed, query)

The issue seems to be how to create the Parsed constant - which
needs the Client to exist for the parsing, but need to be stored
as a constant (because if you send a non-constant to the graphql
library you get an ugly error complaining about dynamic query
errors)

Maybe someone else, even future-me, will be able to resolve this.

### DIFF
--- a/app/controllers/basic_search_controller.rb
+++ b/app/controllers/basic_search_controller.rb
@@ -11,7 +11,23 @@ class BasicSearchController < ApplicationController
     query = QueryBuilder.new(@enhanced_query).query
 
     # builder hands off to wrapper which returns raw results here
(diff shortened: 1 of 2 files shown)


matt-bernhardt:
Closing this as the base branch has merged.

PR #40, opened 2022-05-23
Convert full record view to GraphQL endpoint

matt-bernhardt:

* This is the second part of changing the application to load data from
the GraphQL endpoint rather than REST.

#### Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/rdi-113

* This adds the constant RecordQuery to the Timdex model, in addition to
the already-defined SearcQuery. This is then called from the record
controller, rather than using the older TimdexWrapper model.
* Now unused, the TimdexWrapper model (which wrapped the REST endpoing)
has now been deleted along with its tests and fixtures.
* The record view template has been restructured, because the GraphQL
responds differently than REST does when it errors. There is now a
shared error partial, for example.
* Relevant tests for the TimdexWrapper model have been re-written for
the Timdex model.
* The testing setup is changing slightly:
* Sensitive data is now filtered to the graphql endpoint, rather than
"api/v2"
* Cassettes for the GraphQL endpoint are matched using the request
body in addition to URI and method, as this is where request details
are found.

The first commit on this branch is pretty tightly focused on this ticket, but the second introduces a side effect. As part of this work, I figured out how to use safe operators in the controllers to deal with the entire structure of the API response (go go `dig` method).

Using this knowledge, this fixes some calls in the basic_search controller to use this technique, matching what is happening in the record controller.

#### Developer

#### Code Reviewer

NO

NO


app/views/record/view.html.erb
```diff
@@ -1,63 +1,9 @@
...
-
-<%= field_string(@record, 'source') %>
-
-<%= field_string(@record, 'source_link') %>
-
-<%= field_object(@record, 'subjects') %>
-
-<%= field_list(@record, 'summary') %>
+<% if @record.nil? %>
+	<%= render('record_empty') %>
```
JPrevost, line 6: This seems like odd whitespace. I think the linters run away from the `erb` files so I'm not positive on this.

test/controllers/record_controller_test.rb
```diff
@@ -9,7 +9,8 @@ class RecordControllerTest < ActionDispatch::IntegrationTest
 
   test 'full record path with an id does not return an error' do
     VCR.use_cassette('timdex record sample',
-                     allow_playback_repeats: true) do
+                     allow_playback_repeats: true,
+                     match_requests_on: %i[method uri body]) do
```
JPrevost, line 13: Great call to use this feature to match on the body as well for these now that we've moved to GraphQL. I had a hard time wrapping my head around it this morning, but I'm 100% on board now.

test/test_helper.rb
```diff
@@ -16,9 +16,10 @@
   config.cassette_library_dir = 'test/vcr_cassettes'
   config.hook_into :webmock
   config.allow_http_connections_when_no_cassette = false
-  config.filter_sensitive_data('http://FAKE_TIMDEX_HOST/api/v2/') { ENV.fetch('TIMDEX_BASE').to_s }
+  config.filter_sensitive_data('http://FAKE_TIMDEX_HOST/graphql/') { ENV.fetch('TIMDEX_BASE').to_s }
```
JPrevost, line 19: Can you double check these changes make sense? It seems like `TIMDEX_GRAPHQL` and `TIMDEX_BASE` are being used backward here... but if it works... 🤷🏻

JPrevost (changes requested):
Can you check the README to ensure the new env is documented and old env is removed?

Also, if you don't mind updating the flowchart in the README to remove the QueryWrapper that would be great... although maybe we conceptually still have a wrapper even though we don't call it that? Your call :)

test/test_helper.rb
JPrevost: I'm pretty sure this is working for the scrubbing:

```
  config.filter_sensitive_data('FAKE_TIMDEX_HOST') { ENV.fetch('TIMDEX_HOST').to_s }
  config.filter_sensitive_data('http://FAKE_ORIGIN') { ENV.fetch('TIMDEX_UI_ORIGIN').to_s }
  config.filter_sensitive_data('http://FAKE_TIMDEX_HOST/graphql') { ENV.fetch('TIMDEX_GRAPHQL').to_s }
```

Which we'd then remove TIMDEX_BASE from .env.test

app/views/record/view.html.erb
matt-bernhardt: Oh, yeah - this is tabbed and not spaced. Fixing...

test/test_helper.rb
matt-bernhardt: Yeah, I was having a hard time wrapping my head around how we could trim down ENV with this change - it felt like we definitely could, but was struggling with some details.

matt-bernhardt: Let me take a stab at it now, but if it feels like it's going to drag we can punt to a future ticket yeah

test/models/timdex_test.rb
```diff
@@ -3,7 +3,9 @@
 class TimdexTest < ActiveSupport::TestCase
   def basic_search
     {
-      'q' => 'data'
+      'q' => 'data',
+      'full' => false,
```
JPrevost, line 7: Full is a REST API feature not GraphQL and shouldn't be necessary here. And until pagination is actually enabled I'm not sure it's super helpful to include it here yet but if you have a reason to include it I'm okay with it as I know we'll need it eventually.

app/models/timdex.rb
```diff
@@ -4,14 +4,110 @@
 class Timdex
   HTTP = GraphQL::Client::HTTP.new(ENV.fetch('TIMDEX_GRAPHQL', '')) do
     def headers(*)
-      { 'User-Agent': 'MIT Libraries Client' }
+      {
+        'User-Agent': 'MIT Libraries Client',
+        Origin: ENV.fetch('TIMDEX_UI_ORIGIN', 'http://localhost:3000')
```
JPrevost, line 9: I don't think we should need to set Origin since we are processing the data in Rails and not in the UI directly. I could be wrong, but if you haven't seen an actual need for this in the UI I think removing it removes some potential confusion as to what is going on.

JPrevost (commented):
I'd prefer to remove the Origin and full that I noted, but they aren't really reasons to block. If you'd prefer to land this as-is we can open tickets to pull them back out later.

matt-bernhardt:
Ok, I've pushed two new commits to address those two points (removing TIMDEX_UI_ORIGIN, and stripping out the `full` and `page` parts of the query.

JPrevost (approved):
Looks great. Thanks!

PR #43, opened 2022-05-31
Implement pagination via Pagy gem

matt-bernhardt:

* Many searches will result in more than 20 matching records, but the
  API will only return 20 at a time. As a result, we need an affordance
  in the user interface to allow motivated users to request additional
  batches of results.

#### Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/rdi-112

This implements the Pagy gem to handle the UI's ability to render next/
previous page links. Along the way, other changes are also made:

* The Enhancer and QueryBuilder models now include an awareness of the
  "page" querystring value, which is converted by the query builder into
  a "from" parameter that is passed to Timdex.
* Each of these models can accept unexpected values, although in an
  ideal case the QueryBuilder will not need this protection.
* The SearchQuery within the Timdex model now has support for the "from"
  parameter.

* The QueryBuilder model has a constant named RESULTS_PER_PAGE, set to
  20 because Timdex does not support variable page sizes. If this were
  to change, then it would make sense to store this value in env.
* The Timdex API throws an error if you pass a too-large value for the
  "from" parameter. I have written a test to confirm this, but there
  is probably an improvement to be found for this behavior.
* Some cassettes are regenerated.

#### Developer

#### Code Reviewer

NO

YES - adding Pagy gem


matt-bernhardt:
Closing this in favor of #41

PR #44, opened 2022-06-02
Initial advanced search UI

JPrevost:

* the ability to target specific fields for searches is a core feature

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/RDI-86

* adds a form that accepts values for any field we allow to be targeted
* created advanced_search and search controllers and moves some bits of
  basic_search into search that both search types will share
* Updates query_builder, enhancer, and timdex_search to work with the
  additional query parameters being used for advanced search
* Updates how pagination links are generated to allow for the various
  parameters necessary for advanced searching and filtering

* Pulled updated timdex schema
* Removed display of "notes" from search results (notes remain in the
  query as I believe ASpace records should display a specific type of
  note but defaulting to no notes will be better overall as they are
  mostly weird for a results page)
* Updated readme to correct how to update timdex schame after refactor
  forgot to make that change
* Sources should likely be checkboxes instead of drop-down select but
  TIMDEX can't handle multiple source searching at this time
* Moves some search logic to a shared controller/views for both basic
  and advanced
* Changes display of what was searched for in basic searches in a way
  that allows advanced searching to work as well... however it's not
  super pretty and will need work later

#### Developer

#### Code Reviewer

NO

NO


test/controllers/advanced_search_controller_test.rb
```diff
@@ -0,0 +1,81 @@
+require 'test_helper'
+
+class AdvancedSearchControllerTest < ActionDispatch::IntegrationTest
+  test 'advanced search route exists and renders' do
+    get '/advanced-search'
+    assert_response :success
+  end
+
+  test 'initial form state' do
```
jazairi, line 9: It's worth noting that the criteria here will likely change as we fiddle with the advanced search form, but we can deal with that when we come to it.

jazairi (approved):
Wow, lots more moving parts than I was anticipating. Great work on this!

Part of me wants to clean up the query string, but I don't think it's worth it at this point since it wouldn't really add much beyond visual tidiness. Possibly something to revisit during a slow season.

Thanks for taking this on! :shipit:

PR #45, opened 2022-06-07
Initial external data retrieval for facts

JPrevost:

* Being able to display relevant information for facts we detect in a
  user search is essential as not all records of interests to our users
  will be indexed in our systems

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/RDI-129
* https://mitlibraries.atlassian.net/browse/RDI-130
* https://mitlibraries.atlassian.net/browse/RDI-131
* https://mitlibraries.atlassian.net/browse/RDI-132

* Introduces models for each type of fact that will do external lookups
  relevant to the type of fact, extract appropriate information, and
  then with fact specific cards in the view display the information and
  provide relevant actionable links to users.

* External lookups are triggered in the view synchronously. This is not
  intended to be left this way for long. We _will_ move this to async
  loading soon.
* APIs for external facts are not necessarily the best options and might
  be replaced.
* Date being extracted and displayed is not necessarily the most useful
  for the type of fact we are displaying
* No actionable links are being added to the fact cards (yet)

#### Developer

#### Code Reviewer

NO

NO


jazairi (approved):
Nice work, and thanks for this! Having all of this done up-front will save us a ton of time.

I confirmed that the panels display with the expected metadata for each fact type. Following up on a couple of things we touched on in Slack:

* It makes a little more sense to me to call the facts models in controller methods. I don't think we need to worry about it yet, though, since I assume the logic will shuffle around a bit once we introduce Stimulus.
* We recently had a similar situation to the repeated `fetch` methods that resulted in  refactoring a method to a module: https://github.com/MITLibraries/thing/blob/f1d3d26081e21b7c2dde9c140d49e60dec62c8ec/app/lib/checksums.rb. However, as I reflect on that decision, I'm not so sure it improved the codebase. I thought I should share as an alternative, but I'm satisfied with either approach.

:shipit:

JPrevost:
@jazairi I'd like to introduce a follow-on PR to this that would move to stimulus and address your valid concern about the views triggering data lookups directly. I'll follow up on that in Slack to see if we want to take that on now or wait a bit.

PR #46, opened 2022-06-09
Move fact data loading to async via stimulus 

JPrevost:

* Initial implementation used synchronous loading in the view to
  demonstrate the idea. This was acceptable for demonstration, but it is
  essential for these lookups to not be blocking which this work
  addresses.

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/RDI-133

* Creates a stimulus controller
* Populates a DOM element when a fact is detected that includes a URL
  for stimulus to async call for data which is returned as HTML and
  is added to the DOM when available

- we are testing that the dom element gets the expected data url added
- we are testing what happens when you call that data url
- we are not testing that it all happens at once
- system tests would be required using a headless browser which we will
  discuss and possibly implement depending on how much additional
  complexity it adds to testing versus how much value it provides

#### Developer

#### Code Reviewer

NO

NO


matt-bernhardt (commented):
After looking at this work with VoiceOver, I think a change to the fact panel might be needed. While the markup appears in the DOM without issue and in the expected tab order, I think two changes might be helpful:
1. Add an ARIA role of "live" to the containing panel, so there's some alert to the user when the panel gets populated. This is particularly useful if the API's response is delayed.
2. Make the title of the returned item into a link. Right now, if I tab through the page, the link order jumps from the link to the advanced search form to a link saying "Yes, this" - with no indication of what the statement is referring to.

I'm working through the rest of the change now, but wanted to make this comment now to start the conversation.

app/views/search/_doi.html.erb
```diff
@@ -1,22 +1,9 @@
...
-      <ul>
-        <li><%= json[:publisher] %></li>
-      </ul>
-    </div>
-
-    <div class="panel-footer">
-      <a href="#">Yes this</a> -- <a href="#">Not relevant to my search</a>
-    </div>
-  </div>
+<div id="doi-fact"
```
matt-bernhardt, line 5: It looks like we have repeated `id` values for this structure, between the search partials and the fact templates? Can we remove one, or should we change them to not be identical?

(I'm making the same comment on each of the `doi-fact` elements, but I think all four are doubled)

app/views/fact/doi.html.erb
```diff
@@ -0,0 +1,19 @@
+<% return unless @json.present? %>
+
+<div id="doi-fact" class="fact">
```
matt-bernhardt, line 3: It looks like we have repeated id values for this structure, between the search partials and the fact templates? Can we remove one, or should we change them to not be identical?

(I'm making the same comment on each of the `doi-fact` elements, but I think all four are doubled)

app/views/search/_doi.html.erb
JPrevost: Are you seeing that in the rendered HTML? I thought stimulus was replacing the content in a way that there was only ever one of these ids live at a time... but I may have misinterpreted (or just coded it wrong even interpreting it correctly lol)

JPrevost: Ah, I found it. Definitely doubling up for dois (and likely others). PMID I had checked and it doesn't double because I typod the first ID to be wrong lol. Yup, this needs fixing on all of them. Good catch!

app/javascript/controllers/content_loader_controller.js
```diff
@@ -0,0 +1,15 @@
...
+
+export default class extends Controller {
+  static values = { url: String }
+
+  connect() {
+    this.load()
+  }
+
+  load() {
+    fetch(this.urlValue)
```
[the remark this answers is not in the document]
matt-bernhardt: The more I think about this, I'm not sure that this is a cause for concern - but would love to hear from @jazairi on this point. The "payoff" to breaking through this is that there's another request to the server for a value, but this is possible without the javascript by watching network traffic and hitting whatever internal path is being targeted - something like http://localhost:3000/issn?issn=foo .

As long as I'm not missing a threat here, I think this issue can be ignored.

matt-bernhardt:
I've added a commit to this branch to resolve the two things I think are needed:

1. Adding an `aria-live` value to the hint region, to make sure that its contents get noticed. This _may_ be not quite what this region is meant for, though, so I think we should test this interface with folks who use screenreaders regularly - and test this feature specifically - to make sure that we're in the range of expected behavior.
2. I've removed the duplicate ID values, and slightly simplified markup for the fact panels themselves.

If this looks good to @jazairi , then I'll go ahead and merge.

app/javascript/controllers/content_loader_controller.js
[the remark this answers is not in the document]
jazairi: Yeah, I agree with you both that there seems to be minimal risk here.

jazairi (approved):
This looks good to me. I'd recommend adding a side effect to the rebased commit message that mentions the questions we have about about `aria-live`. Thank you both for your work on this! :shipit:

matt-bernhardt (approved):

PR #48, opened 2022-06-13
Initial refinements to search results UI

jazairi:

This work was prompted by the initial set of UI iterations.

#### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/RDI-99

This proposes a path for the results UI design, focused on the
minimal information needed to identify whether a resource is useful.

Since we will have a full record view, these brief results include
only title, content type, publication date, contributors, and
full-text link. Unlike Bento, they do not include subjects, as I
couldn't discern a use case for this in the brief results. (More on
this in the ticket if you're curious.)

This makes some improvements to a11y, including highlighting
links and adding `:visited` classes. The colors of these are
subject to change; I'm specifically interested in the
reviewer's opinion of the green used for the 'View online' button;
this is borrowed from Bento, but I wonder if it has implications
for color-blind people.

These changes are very much a starting point. I anticipate
additional iteration on the design, and I welcome whatever changes
the reviewer recommends to this current state.

* Adds byebug and updates the gitignore accordingly.
* Contributor types have been removed, but I'm very open to adding
them back in, or possibly changing the number/type of contributors
shown.
* The 'View online' links don't actually go anywhere yet.
* There are some additional features that I think are needed, but
seemed out of scope of this ticket:
    * Search term highlighting;
    * Conditional metadata display for different content types;
    * Full-text links.
* I still have many open questions on this approach, primarily
around the assumptions made re: affordances and metadata. See my
comment in the ticket for more info.

#### Developer

#### Code Reviewer

NO

YES

### DIFF
--- a/app/views/search/_result.html.erb
+++ b/app/views/search/_result.html.erb
@@ -1,40 +1,28 @@
-<li>
+<li class="result">
   <h2 class="record-title">
     <span class="sr">Title: </span><%= link_to(result['title'], record_path(result['timdexRecordId'])) %>
   </h2>
 
-  <% result['contentType']&.each do |type| %>
-    <p>
-      <%= type %>
-    <p>
-  <% end %>
+  <p class="pub-info">
+    <span><%= result['contentType']&.each { |type| type['value'] }.join(' ; ') %></span>
+    <span>
+      <% result['dates']&.each do |date| %>
+        <%= date['value'] if date['kind'] == 'Publication date' %>
+      <% end %>
+    </span>
+  </p>
 
   <% contributors = [] %>
   <% result['contributors']&.each do |contributor| %>
-    <% contributors << "#{contributor['value']} (#{contributor['kind']})" %>
+    <% contributors << contributor['value'] %>
   <% end %>
 
   <p class="result-authors">
     <span class="sr">Authors: </span>
-    <%= contributors.join('; ') %>
+    <%= contributors.uniq.map { |contrib| link_to contrib, results_path({ advanced: true, contributors: contrib }) }.join(' ; ').html_safe %>
   </p>
 
-  <% result['publicationInformation']&.each do |pub| %>
-    <p>
-      <span class="sr">Publication Information: </span>
-      <%= pub %>
-    </p>
-  <% end %>
-
-  <% result['dates']&.each do |date| %>
-    <p>
-      <%= date['kind'] %>: <%= date['value'] %>
-    </p>
-  <% end %>
-
-  <p>
-    Identifier: <%= result['timdexRecordId'] %>
-  </p>
-
-  <hr />
+  <div class="result-get">
+    <%= link_to 'View online', "##{Random.rand(100000)}", class: 'button button-primary green' %>
+  </div>
 </li>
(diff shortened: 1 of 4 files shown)


Gemfile
```diff
@@ -21,6 +21,7 @@ group :production do
 end
 
 group :development, :test do
+  gem 'byebug'
```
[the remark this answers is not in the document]
jazairi: Neat, I didn't realize that was available. Thanks!

app/models/book.rb
```diff
@@ -0,0 +1,10 @@
+class Book
+  def thumbnail
+    return unless @record['pnx']['addata']['isbn']
+
+    # A record can have multiple ISBNs, so we are assuming here that
+    # the thumbnail URL can be constructed from the first occurrence
+    isbn = @record['pnx']['addata']['isbn'].first
+    [ENV.fetch('SYNDETICS_PRIMO_URL', nil), '&isbn=', isbn, '/sc.jpg'].join
```
JPrevost, line 8: This brings up an interesting question... does our license allow us to use those thumbnails for metadata coming from different systems. My assumption is that comes along for usage with Primo/Alma and not "anything we want" but I'm not sure.

JPrevost: I'd suggest we use https://openlibrary.org/dev/docs/api/covers as our initial implementation (and they say they appreciate a link back which we could find a place to include on an "about" page (i.e. they don't expect something intrusive).

jazairi: This was actually included by accident in the latest commit. Something I was starting to think about. 🙂 

You raise a good point, though... if we want thumbnails, we may need to include them in another way.

app/views/search/_result.html.erb
```diff
@@ -1,40 +1,28 @@
...
-  <% result['contentType']&.each do |type| %>
-    <p>
-      <%= type %>
-    <p>
-  <% end %>
+  <p class="pub-info">
+    <span><%= result['contentType']&.each { |type| type['value'] }.join(' ; ') %></span>
+    <span>
+      <% result['dates']&.each do |date| %>
+        <%= date['value'] if date['kind'] == 'Publication date' %>
```
JPrevost, line 10: I'm curious if we'll want to do something with additional dates. For now I agree with this for summary view as a starting point, although once we start considering different types of records I suspect this is an area that will get interesting.

app/views/search/_result.html.erb
```diff
@@ -1,40 +1,28 @@
...
+      <% end %>
+    </span>
+  </p>
 
   <% contributors = [] %>
   <% result['contributors']&.each do |contributor| %>
-    <% contributors << "#{contributor['value']} (#{contributor['kind']})" %>
+    <% contributors << contributor['value'] %>
```
JPrevost, line 17: For some articles, the contributors list will literally be hundreds of people (thanks high-energy physics for including everyone in the field on every article!). We can come back to this later, but something like displaying up to 10 with `et al` on the results view seems safe-ish.

JPrevost (approved):
I like this as an iterative improvement. The Jira ticket points to links to some follow-on work so I'm good with merging this as-is (possibly opening a ticket to investigate abbreviating long lists of authors and considering whether the author search links are useful [as they work now, they are pretty awful but I think that's more of how our search is working than the idea behind them].

app/views/search/_result.html.erb
JPrevost: I don't intend additional work on this ticket on this, I was just thinking out loud.

PR #49, opened 2022-06-17
Combine basic and advanced search UIs

matt-bernhardt:
Combine basic and advanced search UI

* The UI design spike identified that we could probably unify the basic
  and advanced search paths, since they both send traffic to the same
  controller method, and the behavior is compatible.

#### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/RDI-100

* This merges the advanced search form, route, and tests into the basic
  search view. The additional fields are placed inside a details panel,
  which is togglable by keyboard or mouse.
* Closing the details panel - labelled "Clear advanced search fields"
  in the UI - resets those fields to an empty state.
* The validation of the params in the search controller now includes
  a `&.strip` step, as a collapsed search form includes an empty
  advanced field.
* Tests from the advanced_search display are moved over to the
  basic_search controller

Not really a side effect, but a side note - I had hoped to tackle the
search feedback panel as part of this work, but in order to not take
too much time on difficult questions before UX joins us, we have
decided to leave that area alone for now. I have created RDI-163 to
consider the issue of the search feedback area on its own terms.

#### Developer

#### Code Reviewer

NO

NO


test/controllers/basic_search_controller_test.rb
```diff
@@ -6,6 +6,48 @@ class BasicSearchControllerTest < ActionDispatch::IntegrationTest
     assert_response :success
 
     assert_select 'form#basic-search', { count: 1 }
+
```
JPrevost, line 9: Non-blocking: we may want to rename this file and class to `search_controller_test.rb` and `SearchControllerTest` to avoid future confusion.

test/vcr_cassettes/advanced.yml
```diff
@@ -0,0 +1,59 @@
...
+      - max-age=63072000; includeSubDomains
+      Vary:
+      - Origin
+      Content-Length:
+      - '46'
+      Via:
+      - 1.1 vegur
+    body:
+      encoding: UTF-8
+      string: '{"status":500,"error":"Internal Server Error"}'
```
JPrevost, line 57: I think we should have a test with valid responses unless we are intentionally testing error states, which the test doesn't appear to be doing.

test/controllers/basic_search_controller_test.rb
```diff
@@ -6,6 +6,48 @@ class BasicSearchControllerTest < ActionDispatch::IntegrationTest
...
+    get '/?advanced=true'
+
+    assert_response :success
+
+    details_div = assert_select('details#advanced-search-panel')
+    refute_nil details_div.attribute('open')
+  end
+
+  test 'advanced search form appears on results page with URL parameter' do
+    VCR.use_cassette('advanced',
```
JPrevost, line 24: See note in cassette. The cassette returns a 500 error but the test isn't explicitly expecting an error so I'm nervous we could be confused in the future. I'd recommend regenerating the cassette.

test/controllers/basic_search_controller_test.rb
```diff
@@ -6,6 +6,48 @@ class BasicSearchControllerTest < ActionDispatch::IntegrationTest
...
+    assert_nil details_div.attribute('open')
+  end
+
+  test 'advanced search form appears on index with URL parameter' do
+    get '/?advanced=true'
+
+    assert_response :success
+
+    details_div = assert_select('details#advanced-search-panel')
+    refute_nil details_div.attribute('open')
```
JPrevost, line 20: If it works, I might use `assert` instead of `refute_nil` here to avoid the cognitive load of processing double negatives.

test/controllers/basic_search_controller_test.rb
```diff
@@ -6,6 +6,48 @@ class BasicSearchControllerTest < ActionDispatch::IntegrationTest
...
+  test 'advanced search form appears on results page with URL parameter' do
+    VCR.use_cassette('advanced',
+                     allow_playback_repeats: true,
+                     match_requests_on: %i[method uri body]) do
+      get '/results?advanced=true'
+
+      assert_response :success
+
+      details_div = assert_select('details#advanced-search-panel')
+      refute_nil details_div.attribute('open')
```
JPrevost, line 32: If it works, I might use assert instead of refute_nil here to avoid the cognitive load of processing double negatives.

app/views/basic_search/_form.html.erb
```diff
@@ -1,21 +1,118 @@
...
+          NOTE: to allow selecting multiple sources we need to change the way we handle this in TIMDEX to treat source as an
+          "OR" search not "AND"
+        </div>
+      </div>
+    </details>
   </div>
 </form>
 
-<p><%= link_to("Advanced search", advanced_search_path) %>
+<script type="text/javascript">
```
JPrevost, line 83: I'd like to see this inline javascript moved to a new file under `app/javascripts`, but that may be a fair bit of a refactor. If you'd like to hold off on that, please open a ticket to consider a refactor on this.

app/views/search/results.html.erb
```diff
@@ -34,11 +34,7 @@
       </ul>
   </p>
 
-  <% if params[:advanced].present? %>
-    <%= render partial: "advanced_search/form" %>
-  <% else %>
-    <%= render partial: "basic_search/form" %>
-  <% end %>
+  <%= render partial: "basic_search/form" %>
```
JPrevost, line 37: The partial being rendered here should likely move to `search/_form` now that it handles both

JPrevost (changes requested):
There are a couple files that we'd likely be better off renaming to indicate they are handling both basic and advanced searches with this change along with some feedback to consider.

app/views/search/_form.html.erb
```diff
@@ -0,0 +1,83 @@
+<%
+# Conditional input label depends on whether a search has already happened.
```
JPrevost, line 2: This setup block feels like good candidates for a couple of well-named helpers. I'm okay with leaving this here for now though.

JPrevost (approved):
Thanks for the changes. Looks good! 👍🏻 :shipit:

PR #50, opened 2022-06-21
Adds actionable links to DOI fact panel

JPrevost:

* Useful actions in the Fact panel are a large part of the intent of
  fact panels

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/RDI-130

* Uses unpaywall instead of crossref to get more information about
  open availability
* Provides a link to an Open Access copy if one is available
* Always provides a link to Primo/Alma link resolver to check MIT access

* the html/css needs work, but as we want to normalize these fact
  panels I felt we could address that later

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/models/fact_doi.rb
+++ b/app/models/fact_doi.rb
@@ -1,16 +1,30 @@
 class FactDoi
   def info(doi)
-    json = fetch(doi)
-    return if json == 'Error'
+    external_data = fetch(doi)
+    return if external_data == 'Error'
 
+    metadata = extract_metadata(external_data)
+    metadata[:doi] = doi
+    metadata[:link_resolver_url] = link_resolver_url(metadata)
+    metadata
+  end
+
+  def extract_metadata(external_data)
     {
-      title: json['message']['title'],
-      publisher: json['message']['publisher']
+      genre: external_data['genre'],
+      title: external_data['title'],
+      year: external_data['year'],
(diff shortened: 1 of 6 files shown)


app/models/fact_doi.rb
```diff
@@ -1,26 +1,44 @@
...
       JSON.parse(resp.to_s)
     else
-      Rails.logger.debug("Fact lookup error. DOI #{doi} detected but crossref returned no data")
+      Rails.logger.debug("Fact lookup error. DOI #{doi} detected but unpaywall returned no data")
       Rails.logger.debug("URL: #{url(doi)}")
       'Error'
     end
   end
+
+  def open_url(metadata)
```
jazairi, line 41: Non-blocking, but I think OpenURL is one word.

jazairi (approved):
Looks good. I like the use of Unpaywall for this; glad you thought of that! :shipit:

app/models/fact_doi.rb
JPrevost: Good catch. Looking at this with fresh eyes this morning and with that comment in mind, I think it may be better named something more like "primo_url" or "link_resolver_url"

JPrevost: After further discussion on slack, I went with `link_resolver_url`. Thanks for the input!

PR #51, opened 2022-06-23
Add actionable links to ISBN fact panel

jazairi:

The current iteration of this fact panel does not provide
affordances to get the item with the identified ISBN.

#### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/RDI-132

This adds an a tag to the title with the Primo link resolver URL
for the given ISBN.

* This does not include styling updates, as we've decided to
postpone that work.
* Author names have been added to the panel.
* I refactored the actual fetching and response parsing into a
separate method, so as to avoid repetition in the separate `fetch`
methods for ISBNs and authors.
* If we don't have the specific edition that the ISBN refers to,
but we do have other editions of the book, the link resolver isn't
sophisticated enough to indicate this. It might be useful in the
future to indicate in the fact panel whether other editions are
available, but we will likely want UXWS input on this.

#### Developer

#### Code Reviewer

NO

NO


app/views/fact/isbn.html.erb
```diff
@@ -2,13 +2,13 @@
 
 <div id="isbn-fact" class="panel panel-info fact">
   <div class="panel-heading">
-    <%= @json[:title] %>
+    <%= link_to @json[:title], @json[:openurl] %>
```
JPrevost, line 5: Non-blocking idea:

In addition to this title link, we might want to also add (as a replacement for the unimplemented "yes/not relevant links at the bottom of the panel), something like the DOI panel adds for consistency across panels (the specific wording feels weird for ISBN though so maybe `Check MIT Access||Availability`? Not sure.

```ruby
<li><%= link_to('Check MIT Subscription Access', @json[:link_resolver_url]) %>
```

JPrevost (changes requested):

app/views/fact/isbn.html.erb
JPrevost: Yeah I'm not sure we have a clear idea here. I had put those in at some point to demonstrate the idea about feedback so they may come back. I'm totally fine with either direction on this fwiw.

app/views/fact/isbn.html.erb
```diff
@@ -2,17 +2,20 @@
...
   </div>
 
   <div class="panel-body">
     <ul>
-      <li><%= @json[:publish_date] %></li>
-      <li><%= @json[:publishers] %></li>
+      <% if @json[:author_names].any? %>
```
JPrevost, line 10: Tests failed with `.any?`... maybe `.present?` would work here?

JPrevost (approved):

PR #57, opened 2022-07-06
Add highlighted search terms to results

jazairi:
Why these changes are being introduced:

The TIMDEX API now returns search term matches, so we should
display these in the UI.

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/RDI-144
* https://mitlibraries.atlassian.net/browse/RDI-156

How this addresses that need:

This updates the schema and the TimdexSearch model to include the API
changes and adds a highlights partial to render matched terms.
The intent is to highlight term for fields that are not already in the
result view, which is achieved with a couple of helper methods.

It is also possible to highlight matched terms in fields that are
already shown in the results by swapping out the existing field with
the highlight fragment if present. Currently, this would affect
title, pub date, content type, and contributors. I chose not to do
this at this stage because we can't reliably assume that highlight
fragments are the same the original field, partly because they are
limited to 100 characters. We could change in this limit in OpenSearch,
but that would have other implications. This would be a good topic to check
in with UXWS when they can join us on the UI work.

Side effects of this change:

* Due to the changes to the schema, all cassettes that hit TIMDEX
API have been regenerated, and special conditions have been noted
on a couple of tests.
* Currently we are rendering _all_ matches, excluding fields
displayed in the result. Since a single field can have multiple
matches, this can result in a relatively long list of hits for, say,
the 'summary' field. This doesn't seem to be a major issue so far,
but we should check in with UXWS on it.

#### Developer

#### Code Reviewer

NO

NO


.env.test
```diff
@@ -1,2 +1,4 @@
-TIMDEX_HOST=FAKE_TIMDEX_HOST
-TIMDEX_GRAPHQL=https://FAKE_TIMDEX_HOST/graphql
+# TIMDEX_GRAPHQL=http://localhost:3000/graphql
```
JPrevost, line 1: This should be reverted to the non stage values

test/helpers/search_helper_test.rb
```diff
@@ -0,0 +1,19 @@
+require 'test_helper'
+
+class RecordHelperTest < ActionView::TestCase
```
JPrevost, line 3: I think this Class should be `SearchHelperTest`

test/helpers/search_helper_test.rb
```diff
@@ -0,0 +1,19 @@
...
+                               { 'matchedField' => 'dates.value', 'matchedPhrases' => '2022' },
+                               { 'matchedField' => 'contributors.value', 'matchedPhrases' => 'Jane Datascientist' }] }
+    assert_empty trim_highlights(result)
+  end
+
+  test 'does not remove undisplayed fields from highlights' do
+    result = { 'highlight' => [{ 'matchedField' => 'summary', 'matchedPhrases' => 'Have some data' }] }
+    assert_equal [{ 'matchedField' => 'summary', 'matchedPhrases' => 'Have some data' }], trim_highlights(result)
+  end
+end
```
JPrevost, line 19: This is non-blocking, but it might be useful to have a test that shows when the highlights contain both displayed and undisplayed fields, `trim_highlights` returns just the undisplayed fields. That is mostly shown in these tests but since they only cover the two edges, a test for the middle (i.e. both types of fields in the result) would make me feel a little more confident if we make changes to the helper in the future.

JPrevost (changes requested):
Two required changes (env reversion and non-matching class name) and a suggestion for an additional test, but otherwise 👍🏻

test/helpers/search_helper_test.rb
jazairi: Yay for copypasta. Nice catch, thanks. 🙂

PR #64, opened 2022-07-19
Add context about application

jazairi:

We want to provide users with information about the app. This
should be easily configurable, as we are configuring multiple experimental
interfaces under different URLs.

#### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/RDI-213

This adds an 'about' partial that renders the content of the
environment variable `ABOUT_APP`.

* Storing context about the app in env is a bit messy, but it seems to
be the best approach here.
* A new platform (`x86_64-darwin-21`) was added to Gemfile.lock as a
result of running `bundle install`. Documenting here in case it causes
problems.

#### Developer

#### Code Reviewer

NO

NO (but note side effect of new platform added to `Gemfile.lock`)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -302,6 +302,7 @@ GEM
 PLATFORMS
   x86_64-darwin-19
   x86_64-darwin-20
+  x86_64-darwin-21
   x86_64-linux
 
 DEPENDENCIES
--- a/README.md
+++ b/README.md
@@ -49,6 +49,8 @@ change as part of the work.
 
 ## Optional Environment Variables
 
+- `ABOUT_APP`: If populated, an 'about' partial containing the contents of this variable will render on 
+`basic_search#index`.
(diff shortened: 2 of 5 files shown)


jazairi:
@matt-bernhardt As I was working on this, I realized that I'm not sure what the copy should be for the 'about' blurb. Do you have insight on this? I've added the variable with placeholder text to the review apps and staging, but I'm planning to hold off on the 'prod' apps for time time being.

matt-bernhardt (approved):
This looks good to me. I see your note about the content to put in this slot (which IMO also includes whether "About" is the right heading value, or something like "About this application" or "What is this" or ... lots of bikes needing sheds). For now, I'm not sure that I've got a clear idea on that front - and I'm trying to avoid going down what feels like a rabbit hole.

:shipit: and thanks for doing this!

PR #67, opened 2022-07-28
Display complete input values in advanced search form

jazairi:

When the advanced search form is submitted, the full search term
is not retained if it includes a space.

#### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/RDI-237

This wraps advanced search value attributes with quotations, so Rails
knows to retain the full param.

None.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/test/controllers/search_controller_test.rb
+++ b/test/controllers/search_controller_test.rb
@@ -321,6 +321,36 @@ class SearchControllerTest < ActionDispatch::IntegrationTest
     end
   end
 
+  test 'advanced search form retains values with spaces' do
+    VCR.use_cassette('advanced all spaces',
+                     allow_playback_repeats: true,
+                     match_requests_on: %i[method uri body]) do
+      query = {
+        q: 'some data',
+        citation: 'a citation',
+        contributors: 'some contribs',
+        fundingInformation: 'a fund',
+        identifiers: 'some ids',
+        locations: 'some locs',
+        subjects: 'some subs',
+        title: 'a title',
+        advanced: 'true'
+      }.to_query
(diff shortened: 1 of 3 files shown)


test/controllers/search_controller_test.rb
```diff
@@ -321,6 +321,38 @@ class SearchControllerTest < ActionDispatch::IntegrationTest
...
+
+      assert_select 'li', 'Keyword anywhere: some data'
+      assert_select 'li', 'Citation: a citation'
+      assert_select 'li', 'Contributors: some contribs'
+      assert_select 'li', 'Funders: a fund'
+      assert_select 'li', 'Identifiers: some ids'
+      assert_select 'li', 'Locations: some locs'
+      assert_select 'li', 'Subjects: some subs'
+      assert_select 'li', 'Title: a title'
+      assert_select 'li', 'Source: katsu sauce'
```
JPrevost, line 352: I don't think this set of asserts actually tests the fix as I believe would pass on the main branch. To test that the form inputs are populated we'll need to select from something more like `input#advanced-identifiers` etc. (that syntax not tested but something like that... I think.

JPrevost (changes requested):
Code works. Thanks! I think the tests just need a tweak to make sure we are testing the form values not the list above the form.

test/controllers/search_controller_test.rb
jazairi: Ah, you're absolutely right. Thanks for catching that -- will fix.

JPrevost (approved):

PR #69, opened 2022-07-28
Return sourceLink as href for view online links

jazairi:

The 'view online' links in the result view don't currently go
anywhere.

#### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/RDI-238

This assigns the 'view online' hrefs to the sourceLink of each
record. RDI records will always use sourceLink for this purpose,
so this approach is sufficient for the time being.

Future non-RDI use cases may require us to update this logic.

#### Developer

#### Code Reviewer

NO

NO


app/helpers/search_helper.rb
```diff
@@ -8,4 +8,10 @@ def trim_highlights(result)
 
     result['highlight'].reject { |h| displayed_fields.include? h['matchedField'] }
   end
+
+  def view_online(result)
+    return unless result['sourceLink'].present?
+
+    link_to 'View online', result['sourceLink'], class: 'button button-primary green'
```
JPrevost, line 15: I don't think this has been introduced as a change for this PR, but I noticed that `green` is only applied on hover or if I've visited the link. Is that intended? I'm not asking for a change, just curious if we are intentionally having the button "black" for unclicked and "green" for clicked (which seems contrary to what this class applies hence my question).

app/models/timdex_search.rb
```diff
@@ -53,6 +53,7 @@ class TimdexSearch < TimdexBase
             matchedField
             matchedPhrases
           }
+          sourceLink
```
JPrevost, line 56: Such a small change to result in such a mess of cassette updates :)

JPrevost (approved):

app/helpers/search_helper.rb
jazairi: Yes, the intended behavior is that the button's background color is black normally, then green on hover and when visited. The 'green' class comes from our design system; it adds the `a:hover { background-color: green}` rule to our link-buttons instead of the standard `a:hover { background-color: blue }`.

app/models/timdex_search.rb
jazairi: Right? 🤦

PR #75, opened 2022-08-11
Explicitly specify Heroku stack

jazairi:

We do not currently declare the stack we want to use for PR builds.
This is fine as TIMDEX-UI is already using Heroku 22, but it may
be useful for future stack upgrades.

#### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/ENGX-180

This adds an app.json file that declares the target stack and app
name.

None.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app.json
+++ b/app.json
@@ -1,0 +1,4 @@
+{
+  "name": "timdex-ui",
+  "stack": "heroku-22"
+}


app.json
```diff
@@ -0,0 +1,4 @@
+{
+  "name": "timdex-ui",
```
JPrevost, line 2: I'm not sure what `name` does here. It thankfully isn't the name of the app that gets built.

https://devcenter.heroku.com/articles/app-json-schema#name seems to suggest it is just a label, at which point I think this is fine but I'm also not sure if it's noise.

Did you have a specific intent behind adding it? (I'm definitely not opposed, just trying to understand if it does something I don't understand or not)

jazairi: Consistency was the main intent, since we use it in our other apps, but I'm not opposed to removing if you'd prefer.

JPrevost: Nope all good. If it's on the other apps, that's good enough for me.

JPrevost (approved):

PR #79, opened 2022-10-18
Support full records without pub_date or id

JPrevost:

* Some records don't have publication date information (which is expected)

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/TIMX-108

* Adds nil protection for publication dates

* After handling the publication date nil, an error occurred for identifier not being present. I believe this is unexpected and is a result of incomplete Alma mappings in dev1 at this time. Handling nils for identifers seemed fine to include even if the data will likely all include at least one identifier per record.

Sample record to verify the change:
https://timdex-ui-pi-timx-108-p-nalqak.herokuapp.com/record/alma:9935254843506761

If you'd like to see the change locally, make sure you are connected to dev1 graphql and are using the `all-current` alias as your TIMDEX_INDEX and then searching for any alma record should work. Doing the same on the main branch should throw errors.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/helpers/record_helper.rb
+++ b/app/helpers/record_helper.rb
@@ -1,6 +1,6 @@
 module RecordHelper
   def doi(metadata)
-    dois = metadata['identifiers'].select { |id| id['kind'].downcase == 'doi' }
+    dois = metadata['identifiers']&.select { |id| id['kind'].downcase == 'doi' }
     return unless dois.present?
 
     dois.first['value']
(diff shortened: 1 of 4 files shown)


hakbailey:
This looks good from my perspective (I didn't review the code, just verified it works as I would expect). I agree handling nil identifiers is also a good idea since, although we do expect each records to have at least one, they are not technically required.

hakbailey (approved):

jazairi (approved):
Every time something like this comes up, I wonder where else we're missing safe navigation operators...

Thanks for fixing this! :shipit:

PR #111, opened 2024-01-30
Add GDT feature flag

jazairi:

We need a way to toggle GDT features, so we can manage non-GDT TIMDEX UI apps alongside it.

#### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/GDT-157

This adds the `flipflop` gem and a GDT feature.

The `flipflop` gem doesn't appear to be very actively maintained these days. We'll want to keep an eye on that and, if it gets abandoned, consider switching to something like `flipper`.

#### Developer

- [x] All new ENV has been added to Heroku Pipeline, Staging and Prod **(added to `timdex-ui-stage` as well as `timdex-ui-gdt` (prod), since we will likely be using stage for GDT for a while)**

#### Code Reviewer

YES

YES

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,4 @@
-/db/*.sqlite3
-/db/*.sqlite3-journal
-/db/*.sqlite3-[0-9]*
+/db/*.sqlite3*
 /coverage/
 
 # Ignore all logfiles and tempfiles.
--- a/Gemfile
+++ b/Gemfile
@@ -48,3 +48,5 @@ group :test do
   gem 'vcr'
   gem 'webmock'
 end
+
+gem 'flipflop', '~> 2.7'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -135,6 +135,9 @@ GEM
     ffi-compiler (1.0.1)
       ffi (>= 1.0.0)
       rake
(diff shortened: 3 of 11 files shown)


JPrevost (commented):
Hmmmm. Thoughts on moving to `flipper` now proactively rather than sticking with `flipflop`? I know we have experience with `flipflop` so there would potentially be less friction as a reason to stick with it.

jazairi:
Yeah, it's mostly reduced friction. Moving quickly feels important at this stage, and `flipflop` still works. I'd also argue that a move to `flipper` should occur in all our apps, in which case switching TIMDEX UI would be much faster since it would be part of a larger change.

Those are my thoughts. I could be persuaded otherwise.

JPrevost:
@jazairi yeah I was thinking if we tried out `flipper` here, if we liked it we'd update the other apps. However, you are correct in that we have a relatively tight timeline to roll out these core features. Let's focus on what we know now and revisit `flipper` in the (nearish) future.

JPrevost (approved):

jazairi:
@JPrevost I agree that this would be a good test bed for `flipper`. I think it would be worth a spike ticket if we feel like we're ahead of schedule with the UI development.

PR #112, opened 2024-02-05
Redesign filter sidebar

jazairi:

The filter sidebar can become unwieldy, as all available filters are shown at all times.

#### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/GDT-127

This restyles the filter sidebar as dropdown menus. In so doing, it also changes some behavior in the filter logic:

* Number of results is listed in the filter sidebar header.
* Applied filters are highlighted (previously, the 'applied' filter was added, but we hadn't done anything with it).
* Clicking an applied filter removes that filter (this replaces the previous removal mechanism, which removed all applied filters).

* Some tests have been updated to reflect the new logic.
* The sidebar elements have an animation that's tied to height. I've set the height to 200px and added a scroll-behavior rule if the number of filters available exceeds that height. This is somewhat different from how the advanced search sidebar works, so I wanted to draw attention to it unless there are any concerns.
* The JS for the dropdown menus is so minimal that I included it in the view as a `script` tag, but I'm happy to load it via Sprockets if that's preferable.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/views/search/_filter.html.erb
+++ b/app/views/search/_filter.html.erb
@@ -1,19 +1,29 @@
 <% return if values.empty? %>
 
 <div class="category">
-  <h3><%= nice_labels[category] || category %></h3>
-  <ul class="category-terms list-unbulleted">
-  <% values.each do |term| %>
-    <li class="term">
-      <a href="<%= results_path(add_filter(@enhanced_query, category, term['key'])) %>" class="<%= "applied" if @enhanced_query[category.to_sym] == term['key'] %>">
-        <span class="name"><%= term['key'] %></span>
-        <span class="count"><%= term['docCount'] %> <span class="sr">records</span></span>
-      </a>
-    </li>
-  <% end %>
-  </ul>
-  <% if @enhanced_query[category.to_sym].present? %>
-    <div><%= link_to "Show all #{nice_labels[category]&.downcase || category}", results_path(remove_filter(@enhanced_query, category)) %>
-    </div>
-  <% end %>
+  <button class="filter-label <%= 'expanded' if @enhanced_query[category.to_sym].present? || first == true %>"
+          onclick="toggleFilter(this)"><%= nice_labels[category] || category %></button>
+  <div class="filter-options">
+    <ul class="category-terms list-unbulleted">
+    <% values.each do |term| %>
+      <li class="term">
+        <% if filter_applied?(@enhanced_query[category.to_sym], term['key']) %>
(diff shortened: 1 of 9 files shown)


app/views/search/_filter.html.erb
```diff
@@ -1,19 +1,28 @@
...
-  <% if @enhanced_query[category.to_sym].present? %>
-    <div><%= link_to "Show all #{nice_labels[category]&.downcase || category}", results_path(remove_filter(@enhanced_query, category)) %>
-    </div>
-  <% end %>
+  <button class="filter-label <%= 'expanded' if @enhanced_query[category.to_sym].present? %>"
+          onClick="toggleFilter(this)"><%= nice_labels[category] || category %></button>
+  <div class="filter-options">
+    <ul class="category-terms list-unbulleted">
+    <% values.each do |term| %>
+      <li class="term">
```
JPrevost, line 9: I'm not entirely sure why, but the scroll overlaps the content initially in Safari. If I select an element (via tab for example) it reflows the display and looks correct.


Expanded but with nothing selected the scrollbar overlaps the content
Screenshot 2024-02-05 at 2 35 09 PM

Tabbed into an element, the overlap is fixed
Screenshot 2024-02-05 at 3 42 42 PM

app/views/search/results.html.erb
```diff
@@ -64,7 +64,8 @@
     <% if @filters.present? %>
       <aside class="col1q filter-container">
         <div id="filters">
-          <h2>Available filters</h2>
+          <h2 class="hd-3">Filter your results</h2>
+          <h3 class="hd-4"><em><%= @pagination[:hits] %> items</em></h3>
```
JPrevost, line 68: The hits value is a bit confusing for larger sets as displayed here.

Search for "organic chemistry" and it displays `10000 items`. It should probably display something something more like `10,000+ items`. I think we should expect TIMDEX API/OpenSearch to report "more than 10000 items" as "10000" and adjust the UI to anticipate that.

If you want to handle that in a separate ticket that is fine with me.

JPrevost (changes requested):
If we can address the overlapping content before merge that would be great. If that proves hard to track down, I'd be okay with a ticket to loop back on it.

Otherwise this feels like what was requested to me and works with either mouse or keyboard navigation. 👍🏻

jazairi:
The latest commit addresses some initial feedback on this, both from this review and from Darcy (I shared it with her in a meeting yesterday afternoon). There are also a couple of things that came up while I was doing some testing related to the code review/QA feedback. More details in the commit message.

JPrevost (approved):

PR #115, opened 2024-02-09
Add 'Ask GIS' panel and reorganize SCSS

jazairi:
The main purpose of this PR is to add 'Ask GIS' links to all views except the index, as specified in [GDT-129](https://mitlibraries.atlassian.net/browse/GDT-129). This is toggled by a feature flag so it does not appear in other TIMDEX UI apps. (However, a similar feature may be useful for those apps, too.)

Secondarily, this removes some rules from `timdexui.scss`. This is unrelated to the ticket, but I noticed a comment in that file suggesting that those rules should be relocated, and this seemed like a good opportunity to do so.  Check the commit message for details, as two rulesets were removed altogether.

#### Developer

#### Code Reviewer

NO

NO

[GDT-129]: https://mitlibraries.atlassian.net/browse/GDT-129?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/views/record/_sidebar.html.erb
+++ b/app/views/record/_sidebar.html.erb
@@ -8,11 +8,5 @@
     </div>
   <% end %>
 
-  <div class="bit askus">
-    <h3 class="title">Ask Us</h3>
-    <p>Not finding what you're looking for?  We're here to help with anything from quick questions to in-depth research.</p>
-    <p>
-      <a class="btn button-secondary" href="https://libraries.mit.edu/ask/">Ask Us</a>
-    </p>
-  </div>
+  <%= render 'shared/ask' %>
 </div>
--- a/app/assets/stylesheets/partials/_panels.scss
+++ b/app/assets/stylesheets/partials/_panels.scss
@@ -76,3 +76,26 @@
 #advanced-search-panel legend {
   color: #000
 }
+
+.ask-us {
+  margin-top: 3rem;
+  a {
+    margin-top: 0.5rem;
+    &:after {
+      padding-left: 1rem;
+      font-family: FontAwesome;
+      content: '\f075';
+    }
+  }
+  &.view-md {
+    @media (min-width: $bp-screen-md) {
+      display: none;
+    }
+  }
+  &.view-lg {
+    display: none;
+    @media (min-width: $bp-screen-md) {
+      display: block;
+    }
+  }
+}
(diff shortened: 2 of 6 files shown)


jazairi:
Sanity check request for the reviewer: try running the tests locally and let me know if you see that weird behavior I shared in Slack. It seems no longer to be a problem for me with a fresh copy of the repo, for whatever reason.

app/views/record/_sidebar.html.erb
```diff
@@ -8,11 +8,5 @@
     </div>
   <% end %>
 
-  <div class="bit askus">
```
JPrevost, line 11: Aren't we losing this default contact logic the way this is implemented?

I wonder if we could continue down the path you started here and have a single `shared/ask` partial and within that partial split out the gdt/non-gdt logic?

JPrevost:
@jazairi tests worked as expected for me locally

app/views/record/_sidebar.html.erb
jazairi: Yeah, I think that makes sense. The original 'Ask Us' bit felt kind of placeholder-y, but that's not my assumption to make.

jazairi:
@JPrevost Sorry I missed your review on this. I just pushed a commit that I think addresses your feedback. To minimize complexity, I removed the `p` tag in the non-GDT ask us bit and normalized the titles for both. Personally, I prefer the more minimal design that Darcy proposed and figure it should apply to all TIMDEX UI apps, but I don't mind walking that back if you think changing non-GDT stuff is out of scope here.

app/views/record/_sidebar.html.erb
```diff
@@ -8,11 +8,5 @@
...
-    <h3 class="title">Ask Us</h3>
-    <p>Not finding what you're looking for?  We're here to help with anything from quick questions to in-depth research.</p>
-    <p>
-      <a class="btn button-secondary" href="https://libraries.mit.edu/ask/">Ask Us</a>
-    </p>
-  </div>
+  <%= render 'shared/ask_gis' %>
```
JPrevost, line 11: create-review-app

JPrevost, line 11: Can we rename this partial to `shared/ask` now that it is handling both use cases?

jazairi: D'oh. Yes.

JPrevost (approved):

PR #119, opened 2024-02-22
Update pagination according to mockupos

matt-bernhardt:

The GDT project has called for an update to how pagination is shown, across all instances of this app. The details are linked from the ticket below.

### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/gdt-172

* The visual display of the pagination links is changed, including which data points are included. Importantly, we are moving from a display that uses page numbers to one that uses result counts - so there are changes to the analyzer model to reflect this UI.

* We also improve the accessibility of the pagination element by using a nav element, and adding aria labels to make the screen reader output more coherent.

* Tests for both the helper and analyzer are also updated to expect the new markup.

* The analyzer test gets some additional assertions because of the calculations for the first and last record in the overall range.

* The div containing the summary of the current page is still wonky on a screen reader, lacking context in its output. For example, the text `21 - 40 of 95` is read as:

```
21 40 of 95
```
My instinct would be for output like:
```
Showing records 21 to 40 of 95 total
```

However, that would involve either extensive screenreader-only spans (which are read individually, it seems?), or clobbering the whole element with an aria-label, which feels like overkill.

I'd like to ask about this during QA, however, but would encourage the reviewer to listen using VoiceOver to experience it for yourself.

### Environment
- [x] No ENV changes

### Accessibility

### Stakeholder
- [ ] Stakeholder approval has been confirmed
- [x] Stakeholder approval will come during QA (after this merges)

NO

NO

## Code Reviewer

### DIFF
--- a/app/assets/stylesheets/partials/_pagination.scss
+++ b/app/assets/stylesheets/partials/_pagination.scss
@@ -1,15 +1,19 @@
 .pagination-container {
+  clear: both;
+  display: flex;
+  flex-flow: row nowrap;
+  justify-content: center;
   margin-top: 3em;
 
   .previous {
-    text-align: left;
-  }
-
-  .current {
-    text-align: center;
+    border-right: 1px solid black;
+    margin-right: 0.5em;
+    padding-right: 0.5em;
   }
 
   .next {
-    text-align: right;
+    border-left: 1px solid black;
+    margin-left: 0.5em;
+    padding-left: 0.5em;
   }
 }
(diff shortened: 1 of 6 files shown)


matt-bernhardt:
The Codeclimate flag is complaining about the complexity of the initialize method on the analyzer - which now does more calculation directly. I'm open to making the `start` and `end` parameters into separate methods, if you think it will help.

jazairi (approved):
I agree with your screen reader a11y concerns. My instinct is to eschew the dash altogether and have it read as, e.g., `1 to 20 of 100`, but if that's too visually unappealing, then `sr` tags feel viable (albeit a bit messier). I think we should mention this to Darcy during QA and suggest a few options.

Not at all concerned about CodeClimate's complexity concern. It's only a hair above the threshold, anyway.

Thanks for the quick turnaround on this!

PR #120, opened 2024-02-23
Toward replicating fact architecture for searches

matt-bernhardt:
** Why are these changes being introduced:

* The TIMDEX API responses can be quite slow at times, and the response of the application can be frustrating to users when that happens. The browsers' page-loading indications can be subtle, and users now expect more prominent confirmation that their request has been received.

** Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/gdt-143

** How does this address that need:

THIS IS A WORK IN PROGRESS, NOT A FINISHED COMMIT.

* This implements a loading-status message by replacing our search architecture (directly loading results and including them in the first meaningful response to the user) with the approach taken in our fact panels - using a data controller for the search results, while moving the search results to the static page controller.

* The tests are broken, because this is only a partial implementation.

** Document any side effects to this change:

* I am committing these changes and sharing now, because I want a point of comparison with an alternative approach - using turbo frames. The point I've reached here is to now need to replicate the request validation in the static page controller (or to abstract out the validation from the search controller to something called both by search and static controllers). This feels icky, so we are going to see if turbo feels better.

#### Developer

#### Code Reviewer

YES | NO

YES | NO

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -1,6 +1,8 @@
 class SearchController < ApplicationController
   before_action :validate_q!, only: %i[results]
 
+  layout false
+
   def results
     # hand off to Enhancer chain
     @enhanced_query = Enhancer.new(params).enhanced_query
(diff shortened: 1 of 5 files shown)


matt-bernhardt:
This change was supplanted by #122

PR #122, opened 2024-02-26
Make default progress indicator more prominent

matt-bernhardt:

* Our application may already be showing a progress indicator that comes with Rails, but the default rendering is very subtle (only three px high). We want to make this more prominent.

### Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/gdt-143

This makes two changes:

* Sets the threshold for showing the progress to bar to 200 milliseconds instead of the default 500. This will result in users seeing the progress bar more quickly while waiting for a slightly slow page load, which should give them greater assurance that the application is working on a response.

* Once shown, the progress bar is now 75 pixels tall, rather than 3. This is the height of the black navbar at the top of the page, which should make the progress bar much harder to miss when it appears.

This isn't a side effect, but an issue I've become aware of which I can't change at the moment: the default progress bar which Turbo uses is not particularly accessible to screen readers, which means that the "visual indicator of progress happening" is exactly that - purely visual. The screen reader experience while the indicator moves is completely silent.

This is not a new behavior with this change, so it isn't a regression - but it is an example an improvement for one audience not being extended to other audiences. I'm not sure whether this should be enough to block this PR from merging.

The level of effort to make this solution accessible for all audiences would be something like these options:

* Write some javascript to augment the default markup with more accessible markup (aria roles, etc)
* Contribute upstream to turbo-rails to replace the `div` being used by Turbo with more accessible markup
* Look for an existing gem that has already augmented the default markup (I've done some brief searching for such a gem, but nothing immediately jumped out at me)
 
#### ENV
- [x] No ENV needs to change

#### A11y
- [x] The change does not introduce a new accessibility problem, but I
      recognize that it does not address an existing a11y problem. See commit and
      PR text for details.

#### Stakeholders
- [x] Stakeholder approval will happen during QA

NO

NO

### Code Reviewer

### DIFF
--- a/app/assets/stylesheets/application.css.scss
+++ b/app/assets/stylesheets/application.css.scss
@@ -11,6 +11,7 @@
 @import "partials/_local_nav";
 @import "partials/_pagination";
 @import "partials/_panels";
+@import "partials/_progress_bar";
 @import "partials/_search";
 @import "partials/_results";
 @import "partials/_typography";
--- a/app/assets/stylesheets/partials/_progress_bar.scss
+++ b/app/assets/stylesheets/partials/_progress_bar.scss
@@ -1,0 +1,4 @@
+.turbo-progress-bar {
+  background: linear-gradient(to right, #000, #fff);
+  height: 75px;
+}
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -1,3 +1,6 @@
 // Configure your import map in config/importmap.rb. Read more: https://github.com/rails/importmap-rails
 import "@hotwired/turbo-rails"
 import "controllers"
+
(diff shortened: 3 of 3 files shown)


jazairi (approved):
I agree with your accessibility concerns and would support opening a spike ticket to investigate further, or an implementation ticket if you are leaning toward one particular solution. (There is a general [make the app more accessible](https://mitlibraries.atlassian.net/browse/GDT-161 ) ticket that this work may fall under, or perhaps it is a subticket of that.)

Other than that, this is a simple and elegant solution to the problem. Thanks for sleuthing this out!

matt-bernhardt:
Thanks, @jazairi ! I've linked this ticket to that one, and added a note in the comment thread there to remind us to consider this feature at that time.

PR #123, opened 2024-02-28
Add toggle for filters on mobile UI

matt-bernhardt:

The result filters are awkward on a mobile display, and may only be
needed for some users. Rather than having them display by default,
and take up space that is very precious, we want to only show a UI
control to expose them when needed.

### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/gdt-174

* This adds a button above the filters panel markup, and adds a few
classes to the existing markup at the top of the filters panel.

* With CSS, we hide the filters panel on the smallest screens using a
new hidden-sm class. This class is then toggled via javascript when
the user clicks on the UI button.

* The javascript is part of a new file - `app/javascript/filters.js` -
which is loaded by the filters partial.

* I'm not sure where best to add the .hidden-sm class to this codebase.
For now I'm adding it to a new _layouts module.

* During initial testing, it feels awkward to have competing scroll
bars in the filter UI by imposing a maximum height on the panels.
Because the user has now chosen explicitly to show the filters, it
feels better to just have them take up the space they need, rather
than making the user navigate overlapping scroll bars on a linear
display. So this overrides the maximum filter panel height.

### Env
- [x] No ENV implications

### A11y

### Stakeholder
- [x] Stakeholder approval will come during QA

NO

NO

## Code Reviewer

### DIFF
--- a/app/assets/stylesheets/application.css.scss
+++ b/app/assets/stylesheets/application.css.scss
@@ -8,6 +8,7 @@
 @import "partials/_alerts";
 @import "partials/_filters";
 @import "partials/_global_alerts";
+@import "partials/_layouts";
 @import "partials/_local_nav";
 @import "partials/_pagination";
 @import "partials/_panels";
--- a/app/assets/stylesheets/partials/_filters.scss
+++ b/app/assets/stylesheets/partials/_filters.scss
@@ -33,7 +33,7 @@
         scrollbar-gutter: auto;
         scroll-behavior: auto;
 
-        @media (max-width: $bp-screen-sm) {
+        @media (max-width: $bp-screen-md) {
           max-height: inherit;
         }
       }
@@ -123,7 +123,7 @@ button#filter-toggle {
   width: 100%;
   text-align: left;
 
-  @media (min-width: $bp-screen-sm) {
+  @media (min-width: $bp-screen-md) {
     display: none;
   }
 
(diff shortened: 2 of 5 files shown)


jazairi (approved):
Now that we have a `layouts` partial, I wonder if we should move [these rules](https://github.com/MITLibraries/timdex-ui/blob/c570cd3192f6e55757336258903c778d50508c30/app/assets/stylesheets/partials/_panels.scss#L90-L100) to it? If you agree, I can make that change amid another change or two that Darcy has requested to the 'Ask Us' panel.

jazairi (changes requested):
Sorry for the rescinded review. The code looks good, but we actually want this to break at 768px. See the 'results page' section of the[ UI details doc](https://docs.google.com/document/d/1gkz8TtiGnwETp50kVa-LbF4_QNsbRnzw1IiO7w2Nzqs/edit#heading=h.3lqb0z2zj1fv).

matt-bernhardt:
Thanks for catching the needed break point - I've pushed a commit to make this all take effect at the medium screen size, not small.

As for whether we move those other rules into `_layouts.scss`.... that may be logical. I'm not quite clear at the moment, but it will likely look fine during code review - so if you want to include it under that ticket, that seems fine.

jazairi (approved):

PR #126, opened 2024-03-01
Implement location-based search

jazairi:

The GeoData app needs to allow for geospatial searching.

#### Relevant ticket(s):

* [GDT-136](https://mitlibraries.atlassian.net/browse/GDT-136)

This adds form elements for the geospatial fields we’ve added to TIMDEX: geobox and geodistance. Because these fields are relatively complex and all their arguments are non-nullable, each of them has its own drop-down that toggles a boolean param (`:geobox` and `:geodistance`, respectively) to pass the field’s arguments into the TIMDEX query.

Certain client-side form validations have been added (e.g., range for lat/long entries), as well as corresponding controller validations in case a user bypasses the form.

In order to achieve the multiple query permutations (no geo fields, both geo fields, one of either geo field), additional query objects have been added. This is based on the recommendation of the graphql-client gem, which suggests using statically declared heredocs as inputs. However, that gem also recommends making queries out of several smaller heredocs, using GraphQL conveniences like fragments, so our implementation may not be exactly what the gem maintainers intended.

* Introducing an additional monolithic query object for each permutation doesn’t feel scalable or OO. We likely don’t have to time to investigate this now, but it would be interesting to consider alternative methods to constructing queries, and possibly also different GraphQL client gems.
* The controller validations are an important failsafe, but they are not especially friendly. We may want to consider additional client-side validations to reduce the likelihood of encountering the controller validations in the wild.
* Writing this feature revealed that the Enhancer and Query Builder models are functionally identical. This was intentional to prepare the app for additional query enhancements. As we’ve now abstracted that to TACOS, we should consider at some point whether we still need both of these models.
* The form partial and corresponding JS are getting unwieldy. Refactoring these would be a good maintenance activity. Further, the interactive elements are becoming complex enough that E2E testing is worth considering.
* The labels for the new form elements are grotesquely long, but they also need to convey a lot of information. I’m hoping that UXWS can help with this.
* Most cassettes have been regenerated due to schema updates.

#### Developer

#### Code Reviewer

NO

NO

#### Includes updated VCR cassettes?

YES

[GDT-136]: https://mitlibraries.atlassian.net/browse/GDT-136?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jazairi:
Tagging this for review despite CI failure because I'm not sure why CI is failing. The tests pass for me locally, and I'd like the reviewer to try running them locally as a first step.

The codeclimate issues are mostly related to complexity, which is not surprising. I'm open to any suggestions on that front, including breaking down the validation methods further. In many cases, though, it seems to be just above the threshold, in which case I'm less likely to take the alert seriously.

The DRY flagging of the JS code is also relevant, but I'm not sure it's worth fixing. I could be persuaded otherwise.

jazairi:
The codeclimate issues are mostly related to complexity, which is not surprising. I'm open to any suggestions on that front, including breaking down the validation methods further. In many cases, though, it seems to be just above the threshold, in which case I'm less likely to take the alert seriously.

The DRY flagging of the JS code is also relevant, but I'm not sure it's worth fixing. I could be persuaded otherwise.

app/controllers/search_controller.rb
```diff
@@ -60,9 +85,81 @@ def reorder_filters(aggs, active_filters)
...
+    redirect_to root_url
+  end
+
+  def validate_geobox_range!
+    return unless params[:geobox]&.strip == 'true'
+
+    geobox_lat = [params[:geoboxMinLatitude]&.strip.to_f, params[:geoboxMaxLatitude]&.strip.to_f]
+    geobox_long = [params[:geoboxMinLongitude]&.strip.to_f, params[:geoboxMaxLongitude]&.strip.to_f]
+    return if geobox_lat.all? { |lat| lat.between?(-90.0, 90.0) }
+    return if geobox_long.all? { |long| long.between?(-180.0, 180.0) }
```
matt-bernhardt, line 136: Similar to the above, the return clauses here will only fail the validation if both of the latitude and longitude pairs are invalid. A single wrong value out of the four isn't sufficient to fail the query as a whole, because only one wrong value means that the query as a whole contains a valid pair of latitude or longitude values, and that valid pair will trigger the relevant return clause.

test/controllers/search_controller_test.rb
```diff
@@ -455,4 +455,391 @@ def source_filter_count(controller)
...
+      low_query = {
+        geobox: 'true',
+        geoboxMinLongitude: -180.2,
+        geoboxMinLatitude: -90.2,
+        geoboxMaxLongitude: -180.1,
+        geoboxMaxLatitude: -90.1
+      }.to_query
+      get "/results?#{low_query}"
+      assert_response :redirect
+      assert_equal flash[:error], "Latitude must be between -90.0 and 90.0, and longitude must be -180.0 and 180.0."
```
matt-bernhardt, line 702: Something is odd about this test, but I'm having a hard time putting my finger on it. I see that response when I try the query in the UI, but when I try submitting individual out-of-limit values I get different results.

This query works, as it should (although the bounding box is non-sensical, near the south pole)
https://timdex-ui-pi-gdt-136-lo-2ovnii.herokuapp.com/results?q=&geobox=true&geoboxMinLongitude=-180&geoboxMinLatitude=-90&geoboxMaxLongitude=-179&geoboxMaxLatitude=-89

The next two searches each violate only the minimum lat or long parameters, and each seems to respond  with a 500 error from the API, not a flash message:
* minLongitude too low
https://timdex-ui-pi-gdt-136-lo-2ovnii.herokuapp.com/results?q=&geobox=true&geoboxMinLongitude=-180.2&geoboxMinLatitude=-90&geoboxMaxLongitude=-179&geoboxMaxLatitude=-89

* minLatitude too low
https://timdex-ui-pi-gdt-136-lo-2ovnii.herokuapp.com/results?q=&geobox=true&geoboxMinLongitude=-180&geoboxMinLatitude=-90.2&geoboxMaxLongitude=-179&geoboxMaxLatitude=-89

The next two searches violate only the _maximum_ lat or long parameters, and appear to be tripped up by the bounding box checks for whether min and max values are in the right order? The flash message reads "Bounding box minimum lat/long cannot exceed maximum lat/long"
* maxLongitude too low
https://timdex-ui-pi-gdt-136-lo-2ovnii.herokuapp.com/results?q=&geobox=true&geoboxMinLongitude=-180&geoboxMinLatitude=-90&geoboxMaxLongitude=-180.2&geoboxMaxLatitude=-90

* maxLatitude too low
https://timdex-ui-pi-gdt-136-lo-2ovnii.herokuapp.com/results?q=&geobox=true&geoboxMinLongitude=-180&geoboxMinLatitude=-90&geoboxMaxLongitude=-179&geoboxMaxLatitude=-90.2

As I read the search controller, it looks like the checks should be performed in the order of "presence, range, values" - so I would expect that the above geobox searches would trip the range validations first, and not reach the value validations - but that's not what is happening (and in some cases the invalid values are getting passed by every check, and the API is returning an error).

Looking at the geobox range validation again, I think there's at least one issue with the logic (which I've commented up there). I wonder whether the testing strategy here should be to send in one invalid value at a time, rather than a set of entirely invalid values?

test/controllers/search_controller_test.rb
```diff
@@ -455,4 +455,391 @@ def source_filter_count(controller)
...
+    end
+
+    test 'geobox lat/long cannot exceed max limit' do
```
matt-bernhardt, line 705: I think the same issues with the minimum-value test are afflicting this test as well.

Valid, no errors:
https://timdex-ui-pi-gdt-136-lo-2ovnii.herokuapp.com/results?q=&geobox=true&geoboxMinLongitude=179&geoboxMinLatitude=89&geoboxMaxLongitude=180&geoboxMaxLatitude=90

Bumping either of the maximum values beyond the limit results in a 500 error, not a validation error.

EDIT - I think I found the cause - the validation clause being tested here requires both pairs of latitudes and longitudes to include an invalid value in order to return a failure.

test/controllers/search_controller_test.rb
```diff
@@ -455,4 +455,391 @@ def source_filter_count(controller)
...
+        geoboxMinLatitude: 80.0,
+        geoboxMaxLongitude: 29.6,
+        geoboxMaxLatitude: 56.9
+      }.to_query
+      get "/results?#{bad_values}"
+      assert_response :redirect
+      assert_equal flash[:error], "Bounding box minimum lat/long cannot exceed maximum lat/long."
+    end
+
+    test 'geodistance lat/long cannot fall below min limit' do
```
matt-bernhardt, line 765: I'm able to evade these validations by passing only one invalid value, and not two:
Works
https://timdex-ui-pi-gdt-136-lo-2ovnii.herokuapp.com/results?q=&geodistance=true&geodistanceLatitude=90&geodistanceLongitude=180&geodistanceDistance=50mi

Invalid longitude
https://timdex-ui-pi-gdt-136-lo-2ovnii.herokuapp.com/results?q=&geodistance=true&geodistanceLatitude=90&geodistanceLongitude=180.1&geodistanceDistance=50mi

Invalid latitude
https://timdex-ui-pi-gdt-136-lo-2ovnii.herokuapp.com/results?q=&geodistance=true&geodistanceLatitude=90.1&geodistanceLongitude=180&geodistanceDistance=50mi

Looking again at the validation clause, I tried to note above what I think the problem is.

test/controllers/search_controller_test.rb
```diff
@@ -455,4 +455,391 @@ def source_filter_count(controller)
...
+        geodistanceLatitude: -90.1,
+        geodistanceLongitude: -180.1,
+        geodistanceDistance: '50mi' 
+      }.to_query
+      get "/results?#{low_query}"
+      assert_response :redirect
+      assert_equal flash[:error], "Latitude must be between -90.0 and 90.0, and longitude must be -180.0 and 180.0."
+    end
+
+    test 'geodistance lat/long cannot exceed max limit' do
```
matt-bernhardt, line 777: Same as above - this test fails because both the latitude and longitude values are incorrect. Passing only one wrong value causes the validation to pass incorrectly.

test/controllers/search_controller_test.rb
```diff
@@ -455,4 +455,391 @@ def source_filter_count(controller)
...
+        geodistanceLatitude: 90.1,
+        geodistanceLongitude: 180.1,
+        geodistanceDistance: '50mi' 
+      }.to_query
+      get "/results?#{high_query}"
+      assert_response :redirect
+      assert_equal flash[:error], "Latitude must be between -90.0 and 90.0, and longitude must be -180.0 and 180.0."
+    end
+
+    test 'geodistance min range limit' do
```
matt-bernhardt, line 789: It took me a bit to understand what this test is doing - am I correct in understanding that this is testing that the lat/long values can go right up to the edge of the allowable envelope? (i.e. the validations don't kick in _at_ 90 or 180, but _after_ 90 or 190). I'm not sure what to do about this - I can't think of a better name, but that's what I find myself wanting - to make clear what is being tested here.

matt-bernhardt (changes requested):
This is a very complex feature, you're not kidding. I particularly agree that the form javascript feels like it should be refactored, but I don't think it should be done here. It feels more like a maintenance days project.

There's a few things in here which I think need to be adjusted. I've tried to provide examples where relevant - but I think the summary is:

* The validation clauses in the search controller have bugs in them that prevent them failing when a single incorrect value is submitted. I think a status variable in the clauses (start as true, either check failing sets the status to false, and then return only if status is still true) might be what's needed. I'd also recommend writing tests that confirm this by submitting only a single invalid value, rather than entire sets of incorrect values.

* The UI validations being applied to latitude and longitude aren't consistent between the geobox and geodistance searches. Geobox is too permissive with latitude, while geodistance is too restrictive with longitude.

* I don't think the ordering requirement in geobox longitudes should be imposed, as I believe there are valid bounding boxes which violate that constraint (for one example, a bounding box for the Pacific Ocean would have a "minimum" longitude of ~105 to get to Vietnam, with a "maximum" longitude of -69 to get all the way to Chile). We've committed to the terms "minimum" and "maximum", but I wonder in hindsight whether a better choice would have been "left" and "right".

* There's a comment about the validation on the geodistanceDistance parameter which is non-blocking, and also a thought about test naming for when we are poking on the equal-is-valid nature of these constraints. I don't think those are blocking, however.

* I haven't looked at all of the cassettes, but I see that the test suite as a whole is passing, so I think that's sufficient?

* My eyes are crossing a little bit right now, so I might want to look at the javascript file again to see if anything jumps out at me. It hasn't so far in my poking on the application locally, however.

Despite all of this, thank you very much for your work on this feature. I have a better understanding now of what you've been grappling with to make this all happen, and I don't want my comments to make it seem like I don't recognize the level of effort you've put in so far on this.

app/controllers/search_controller.rb
```diff
@@ -60,9 +85,81 @@ def reorder_filters(aggs, active_filters)
...
+
+  def validate_geobox_values!
+    return unless params[:geobox]&.strip == 'true'
+
+    geobox_params = [params[:geoboxMinLatitude]&.strip.to_f, params[:geoboxMinLongitude]&.strip.to_f,
+                     params[:geoboxMaxLatitude]&.strip.to_f, params[:geoboxMaxLongitude]&.strip.to_f]
+
+    # Confirm that min values are lower than max values
+    return if geobox_params.any?(&:blank?)
+    return if geobox_params[0] < geobox_params[2] && geobox_params[1] < geobox_params[3]
```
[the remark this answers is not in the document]
jazairi: That's kind of a relief. This is one of the trickier validations, and I think it would become especially complex when we start adding more of these validations to client-side.

app/controllers/search_controller.rb
```diff
@@ -60,9 +85,81 @@ def reorder_filters(aggs, active_filters)
...
+    geobox_lat = [params[:geoboxMinLatitude]&.strip.to_f, params[:geoboxMaxLatitude]&.strip.to_f]
+    geobox_long = [params[:geoboxMinLongitude]&.strip.to_f, params[:geoboxMaxLongitude]&.strip.to_f]
+    return if geobox_lat.all? { |lat| lat.between?(-90.0, 90.0) }
+    return if geobox_long.all? { |long| long.between?(-180.0, 180.0) }
+
+    flash[:error] = 'Latitude must be between -90.0 and 90.0, and longitude must be -180.0 and 180.0.'
+    redirect_to root_url
+  end
+
+  def validate_geodistance_values!
```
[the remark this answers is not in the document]
jazairi: Yeah, validating distance units (or lack thereof) is something I left alone for me. In the API documentation, it's a suggestion, but not a requirement, and I don't know what the default is (probably km?) or how it's parsed. A better understanding of the latter would make it easier to validate; i.e., we could check for the presence of some common units (maybe 'mi' and 'km' to start). I'll dig into the OpenSearch side and return to this in a bit.

app/assets/stylesheets/partials/_search.scss
```diff
@@ -64,6 +66,19 @@
     }
   }
 
+  #geobox-search-panel,
+  #geodistance-search-panel {
+    fieldset {
+      legend {
+        color: $black;
```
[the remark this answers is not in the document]
jazairi: Absolutely! To be honest, I didn't know that other rule existed. The style guide seems to have overwritten it, somehow, as the legend color before adding this rule was the style guide default (one of the grays), which caused a contrast issue.

test/controllers/search_controller_test.rb
jazairi: Yeah, I found the same issue when you mentioned the 500 errors in Slack. The current range validation logic can prevent it from erroring if it returns before it finds an invalid value.

test/controllers/search_controller_test.rb
jazairi: Correct, the intent here is to test the minimum (and, in the other test, the maximum) allowable values. I'll retitle these tests to make that clearer.

jazairi:
Thanks for the detailed review, Matt!! I'll get to work on these changes momentarily.

jazairi:
@matt-bernhardt I _think_ this catches everything, but it's quite likely I missed something. Please let me know either way.

The changes to the form are unrelated to your initial review:
* Placeholders are now sample values. The downside to this is that we are not providing the same richness of information to all users, as not all screen readers process placeholders. I plan to raise this issue during UX review.
* The instructions for each field are now in a span below the input. This allows us to assign `aria-describedby` and makes the fieldsets a bit less cluttered to my eye. (Incidentally, this would also be a relocation candidate for the placeholder text.)

Validation of distance units wasn't specifically requested, either, but your comment made me think it would be useful to handle that 500 error you were seeing with invalid units. You'll note that in the flash error, I didn't technically list _all_ of the possible units, as it seemed easier to encourage users to abbreviate them. (Part of this is due to some oddness in the OpenSearch conventions; if I were searching in inches, I would not think to use `40inch`).

Thanks again for the in-depth review!

app/views/search/_form.html.erb
```diff
@@ -100,37 +112,42 @@ end
...
-                     id="geodistance-longitude" name="geodistanceLongitude" placeholder="Longitude"
-                     value="<%= params[:geodistanceLongitude] %>"
-                     <%= 'required="required" aria-required="true"' if geodistance_required %>>
+              <label class="geo-label" for="geodistance-longitude">Longitude</label>
+                <input type="number" step="0.000001" min="-180.0" max="180.0"
+                       class="field field-text <%= "required" if geodistance_required %>" 
+                       id="geodistance-longitude" name="geodistanceLongitude"
+                       value="<%= params[:geodistanceLongitude] %>" aria-describedby="long-desc"
+                       <%= 'required="required" aria-required="true"' if geodistance_required %>
+                       aria-describedby="long-desc" placeholder="E.g., -83.732124">
```
matt-bernhardt, line 136: I see what you did with those lat and long values. Well played :-)

matt-bernhardt (changes requested):
I think this all looks good to me, with the lone exception of the minimum longitude field under geobox having an incorrect description value (the input itself is correct, but described wrong).

app/views/search/_form.html.erb
jazairi: 😈

app/views/search/_form.html.erb
```diff
@@ -52,44 +52,56 @@ end
...
+              <input type="number" step="0.000001" min="-180.0" max="180.0"
                      class="field field-text <%= "required" if geobox_required %>" 
-                     id="geobox-minLongitude" name="geoboxMinLongitude" placeholder="Minimum longitude"
-                     value="<%= params[:geoboxMinLongitude] %>"
-                     <%= 'required="required" aria-required="true"' if geobox_required %>>
+                     id="geobox-minLongitude" name="geoboxMinLongitude" value="<%= params[:geoboxMinLongitude] %>"
+                     <%= 'required="required" aria-required="true"' if geobox_required %>
+                     aria-describedby="minLong-desc" placeholder="E.g., -73.507239">
+              <span class="geo-desc" id="minLong-desc">
+                A decimal between -90.0 and 90.0 (Western hemisphere is negative)
```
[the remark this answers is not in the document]
jazairi: Ugh, more copypasta typos. Sorry about that.

matt-bernhardt (approved):
:shipit:

PR #128, opened 2024-03-04
Expand available filters, give applications the ability to filter and order via ENV

matt-bernhardt:
This expands the set of available filter / aggregation categories that are available for our search applications, and makes it possible for each application to customize which categories are shown, and how.

With this change, the full list of available categories are:
* Content type
* Contributor
* Format
* Language
* Literary form
* Source
* Subject

Individual applications can, via an `ACTIVE_FILTERS` environment variable, choose which of these are shown, and in what order.

Additionally, applications can override the name of any of these categories using a separate environment variable for each category.

This work will partially address ticket [GDT-128](https://mitlibraries.atlassian.net/browse/GDT-128), but it cannot close until the following filters are also supported:

* Access type
* Place (this is likely ready to go, I'm just wanting to focus on the first seven for now)
* Publisher
* Year

### Environment

### Accessibility
- [ ] This change does not affect the view layer, so far as I can determine

### Stakeholders
- [ ] Stakeholder approval will come during QA

NO

NO

## Code Reviewer

[GDT-128]: https://mitlibraries.atlassian.net/browse/GDT-128?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jazairi (approved):
Nice work on this! This will make it much easier to add the other filters when they become available. Abstracting the config to env is a nice touch that will help us manage multiple TIMDEX apps with different filters and labels.

I agree both that it would be tiresome to test everything in `nice_labels`, and that it opens us up to missing a potential breaking change. In this case, I think the logic in that method is so simple that testing one filter as a sample is fine for now. As you note, it's written to fall back on reasonable labels, anyway. It seems to be that the absolute worst case is that a new filter label is not defined at all, in which case it will fall back to the GraphQL field name, which doesn't feel particularly dangerous.

A minimally painful way to test everything could be to just add all the label to those two tests, so if something is off with anything in the `nice_labels` hash, the test fails and alerts us to the issue. That may be something to consider we've added the last of the filters. For now, this feels like adequate coverage.

Thanks for the excellent work, and for going above and beyond the requirements!

jazairi:
One more thing to add, in case you mention this detail in an amended commit message: GDT-128 can actually merge *without* the access type filter. That one has [its own ticket (GDT-140)](https://mitlibraries.atlassian.net/browse/GDT-140), as it's a bit different from the others.

PR #130, opened 2024-03-06
Add GDT access links

jazairi:

The GDT UI requires various access links to appear in the record sidebar. The text and hrefs of these depend on the data provider and access restrictions.

#### Relevant ticket(s):

* [GDT-141](https://mitlibraries.atlassian.net/browse/GDT-141)
* [GDT-175](https://mitlibraries.atlassian.net/browse/GDT-175)
* [GDT-176](https://mitlibraries.atlassian.net/browse/GDT-176)

This applies the requested changes:

* A 'Download geodata files' link appears for MIT-provided records that are OA.
* A 'Download geodata files' link with an auth notice appears for MIT-provided records that require auth.
* A 'View `[provider]` record' link appears for non-MIT records, where `[provider]` is the institution providing access to the data.

Non-MIT records always use the `kind=Website` link, whereas MIT records always use the `kind=Download` and `text=Data` link.

GDT-175 is *not* implemented in this commit, as we have not yet loaded GIS records from Alma.

* The schema has been updated, and some unrelated cassettes have been regenerated as a result.
* The `provider` field has been added to the `Record` query object.
* A `record` SCSS partial has been added.
* The controller tests rely on specific data returned by the API call. We've generally avoided this testing approach as it can be brittle, but this seemed like the best way to test the conditional logic in the new `access_button` partial. We do have tests covering the helper methods, so we could consider removing these tests for now. Either way, we should evaluate E2E testing for this repo (and likely others).

#### Developer

#### Code Reviewer

NO

NO

#### Schema updated?

YES

[GDT-141]: https://mitlibraries.atlassian.net/browse/GDT-141?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[GDT-175]: https://mitlibraries.atlassian.net/browse/GDT-175?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[GDT-176]: https://mitlibraries.atlassian.net/browse/GDT-176?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


test/helpers/record_helper_test.rb
```diff
@@ -101,4 +101,149 @@ class RecordHelperTest < ActionView::TestCase
...
+                    ]
+                  }
+    access_auth = {
+                    'rights' => [
+                      {
+                        'description' => 'foo',
+                        'kind' => 'bar'
+                      },
+                      {
+                        'description' => 'Restricted',
```
matt-bernhardt, line 229: This may be too finicky a comment, in which case I apologize for being pedantic. I'm trying to make sure that I understand the logic of the helper method, and I'm not sure whether a record like this will actually end up generating an authenticated button, or an external link.

The check in the access_button partial appears to be looking for the access_type helper to return `MIT Authentication` as a value, rather than `Restricted`?

I also realize that this test is only testing the `gis_access_link` helper, and that we aren't trying to test the view results. So maybe this is not something that needs to be adjusted. I bring it up only because I think this is what was proving tricky for me to understand as I tried to follow the helper logic.

matt-bernhardt (approved):
None of my comments are blocking, and I may just be borrowing trouble - so I'm happy to approve this as-is.

One comment is trying to anticipate the future of a possible conflict or gap between this and the geosearch PR.

The other PR asks about whether one of the fixtures in the tests should be tweaked, or if maybe I've misunderstood the flow of the helper methods - it feels like the view templates would be expecting a value that the fixture doesn't use, but which may not be relevant because we're not testing the view layer (nor should we be - but IMO there is value in using realistic data structures where possible).

If all this is irrelevant, please ignore me and merge as-is.

app/models/timdex_record.rb
```diff
@@ -69,6 +69,7 @@ class TimdexRecord < TimdexBase
         }
         numbering
         physicalDescription
+        provider
```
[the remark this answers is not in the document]
jazairi: Yes, that's correct. It sucks, but I suspect that refactoring the logic will suck more. 🙂 Refactoring everything into fragments seems a bit tricky, and while I think it's a good idea, it feels very post-MVP to me given the current timeline. In fact, I think part of that change would involve a spike ticket evaluating whether we even want to use the `graphql-client` gem at all, since we're not really using any of its features.

test/helpers/record_helper_test.rb
jazairi: Yes, that's a good point. I think this was an oversight on my part. While the test outcome isn't affected by the access type in this case, it's a good idea to keep them consistent.

matt-bernhardt (approved):
Thanks - looking forward to seeing these in place!

I agree, also, about the prospect of refactoring the query in the record model. This feels like the sort of somewhat-maintenance-but-also-not that we need to make time for, but maybe not right now.

:shipit:

PR #131, opened 2024-03-07
Make new aggregations searchable

jazairi:
Why these changes are being introduced:

We added some more aggregations as part of the ongoing work, but did
not add them to the query string or corresponding models.

Relevant ticket(s):

* (GDT-128)[https://mitlibraries.atlassian.net/browse/GDT-128]

How this addresses that need:

This adds the new aggregations to the Enhancer and QueryBuilder models, and adds them to the query in the TimdexSearch model.

Side effects of this change:

* VCR cassettes have been regenerated.
* This is another instance of something that has come up recently: the need to develop an E2E testing strategy for our Rails apps, particularly those that are API consumers. We can test this using our standard VCR process, but that is likely to break when cassettes are regenerated and the search results change.

#### Developer

#### Code Reviewer

NO

NO


jazairi:
I'm treating this as a hotfix and merging it prior to review. I welcome a retroactive review and any suggestions/opinions on the testing front.

app/views/search/_filter.html.erb
```diff
@@ -1,17 +1,17 @@
 <% return if values.blank? %>
 
 <div class="category">
-  <button class="filter-label <%= 'expanded' if @enhanced_query[category.to_sym].present? || first == true %>"
+  <button class="filter-label <%= 'expanded' if @enhanced_query[category].present? || first == true %>"
```
matt-bernhardt, line 4: I see us dropping the `.to_sym` method here, and I'm not sure why. I'm sure it makes sense - my first assumption is because the values aren't convertible to symbols now, but were before. I'm just not sure I'm following what changed.

matt-bernhardt (commented):
There's one change in here that I'm not sure I'm following (not because I anticipate a problem, I'm just not sure why it is happening) - I've noted that inline.

More generally, am I correct in thinking that these are things I should have included as part of #128 ?

app/views/search/_filter.html.erb
jazairi: Sorry, I should've been clearer in the commit message. That `to_sym` call was there because `SearchController#extract_filters` didn't always cast the filter keys as symbols. I refactored that method to do so a while back, but noticed while debugging this that there were still a couple of `to_sym` calls. I removed them to make this less confusing for our future selves (hopefully), but it's unrelated to the additional/customizable filters changeset.

jazairi:
@matt-bernhardt Yes, this is related to #128. I ought to have mentioned it during code review but missed it at the time. I suspect that I was confirming functionality by click-testing `contentTypeFilter` and `sourceFilter`, which had already been set up. Regardless, something I should've caught as reviewer; my mistake!

app/views/search/_filter.html.erb
matt-bernhardt: Ah, gotcha - that makes sense. Thanks!

matt-bernhardt:
No worries - I should have caught it as well. Thank you for catching it now, and for the hotfix :-)

PR #133, opened 2024-03-14
Update result view for GDT

jazairi:

The GeoData app requires specific metadata in the results view.

#### Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/GDT-130

This adds a separate result view for geo records that includes the specifed metadata fields. It also applies some general styling updates, most of which involve spacing and font size/weight. The most significant CSS change is that values that may be long (authors, higlights, summary) are now truncated.

* The date parsing logic catches a lot of edge cases, but not all of them. It also puts a fair amount of faith in the source data to have reasonably formatted dates. Finally, it makes the assumption that the date we want is coverage, falling back to issued. This is subject to change.
* Certain fields have been added to support these changes, and cassettes have been regenerated as a result.
* I noticed an indentation issue in the search controller test, so
some lines that appear to have changed have just been re-indented.
New tests begin at line 1036.
* Certain fields have been added to support these changes, and
cassettes have been regenerated as a result.

#### Developer

#### Code Reviewer

NO

NO

#### Includes regenerated cassettes?

YES

#### Includes query object updates?

YES


app/helpers/search_helper.rb
```diff
@@ -14,4 +14,56 @@ def view_online(result)
 
     link_to 'View online', result['sourceLink'], class: 'button button-primary green'
   end
+
+  def view_record(record_id)
+    link_to 'View full record', record_path(id: record_id), class: 'button button-primary green'
+  end
+
+  def number_in_results(result)
```
JPrevost, line 22: This appears to be adding numbers to an unordered list. I wonder if we could instead use ordered list and update the `start` attribute calculated instead of this type of calculation.

If `ol` doesn't make sense, I wonder if we can move this into the main results loop instead of looping all the results inside the already looping results loop.

app/helpers/search_helper.rb
```diff
@@ -14,4 +14,56 @@ def view_online(result)
...
+    # Ignore dates of the same kind with unique values.
+    return if relevant_dates.length > 1
+
+    relevant_date = relevant_dates.first['value']
+
+    # If the date vaguely resembes a year, return it as is.
+    return relevant_date if relevant_date.length == 4
+
+    # If the date vaguely resembles 'YYYY-MM', 12/01/2020, or another unparsable date, extract the year.
+    handle_unparsable_date(relevant_date)
```
JPrevost, line 49: I wonder if we can rely on some built-ins to coerce these strings into dates and then extract the years.

app/views/search/results.html.erb
```diff
@@ -39,7 +39,11 @@
     <div class="col3q wrap-results">
       <% if @results.present? %>
         <ul id="results" class="list-unbulleted">
-          <%= render(partial: 'search/result', collection: @results) %>
+          <% if Flipflop.enabled?(:gdt) %>
```
JPrevost, line 42: I have moderate concerns about forking `results` entirely rather than determining which features are geo specific and which are appropriate for all timdex-uis. The concern stems from the fact we'll eventually need to figure this out and I'm not sure if we'll ever have more insight into the differences/similarities than we do now.

This isn't blocking, but it is requesting a bit of context as to why this path was chosen.

app/helpers/search_helper.rb
JPrevost: Alternatively, I wonder if our GraphQL should have a field to return the `as_date` version of this field.

JPrevost (commented):
I think I'd like to chat or otherwise hear from you on what the reasoning was to fork the results list for geo records rather than trying to integrate changes with the main list. I suspect you have good reason, but it isn't entirely clear to me that this was the correct path. Slack/zoom may be good to chat about this as it feels like a complicated thing to work out in code review.

app/helpers/search_helper.rb
```diff
@@ -14,4 +14,56 @@ def view_online(result)
...
+    result_index = 0
+    @results.each_with_index do |record, record_index|
+      result_index = record_index if record['timdexRecordId'] == result['timdexRecordId']
+    end
+    result_index + @pagination[:start]
+  end
+
+  # 'Coverage' and 'issued' seem to be the most prevalent types; 'coverage' is typically formatted as
+  # 'YYYY', whereas 'issued' comes in a variety of formats.
+  def parse_geo_dates(dates)
```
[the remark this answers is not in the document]
jazairi: Fairly confident. I suspect most print resources would use `kind=Publication date`, for example.

app/helpers/search_helper.rb
jazairi: `Date.strptime` provides some convenience, but you still have to infer the input format somehow in order to know how to parse it.

app/views/search/results.html.erb
jazairi: They're similar enough that I did consider using the same partial with a handful of conditions. This decision was informed partly by my preference for smaller partials, and partly by my assumption that this approach would be preferable for the record view. (The logic would get more complex in that case, and I think more thought needs to be given to what fields we want in non-GDT records before we assume that they should be close to identical.)

I also have questions about some questions about the certain details that I don't think we can answer yet (e.g., whether 'summary' be in the result view for all record types). But, I understand your point that we may not gain further insight into that for a long time. I also think it would be easier to reconcile the differences between GDT and non-GDT views if we throw them both into the same file. That goes for the record view, too: while a single partial will be messier now, it could make things easier on us later.

I'm fairly ambivalent on this one, and happy to go the opposite route if you have a strong preference.

app/helpers/search_helper.rb
```diff
@@ -14,4 +14,56 @@ def view_online(result)
...
+    relevant_dates = if dates&.any? { |date| date['kind'] == 'Issued' }
+                       dates.select { |date| date['kind'] == 'Issued' }&.uniq
+                      elsif dates&.any? { |date| date['kind'] == 'Coverage' }
+                       dates.select { |date| date['kind'] == 'Coverage' }&.uniq
+                     end
+    return if relevant_dates.blank?
+
+    # Ignore dates of the same kind with unique values.
+    return if relevant_dates.length > 1
```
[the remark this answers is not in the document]
jazairi: Yeah, I'm pretty 🤷 with this, but I do think you're right that we should probably return something rather than nothing.

app/views/search/results.html.erb
JPrevost: Thanks for the context. I'm okay with going with your decision here to use the separate partials to make each partial cleaner and deal with this all later. As the only source that has gotten actual attention is GIS, I've come around to agreeing that we'll need more eyes on the non-GIS records to not just be guessing as to what should be the same.

app/helpers/search_helper.rb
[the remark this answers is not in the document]
JPrevost: I guess maybe it comes down to... is it better to display nothing or display something that is only a "partial truth". i.e. displaying 1999 when in reality it should say something like 1999/2005.

If we leave the current logic, let's update the comment to note that this might need further reflection and why.

JPrevost (approved):
If you'd like to try a refactor to OL or make any other changes that's fine. If you'd prefer to backlog the change to OL that's fine too.

JPrevost (approved):

PR #135, opened 2024-03-15
Update record view for GDT

jazairi:

The GeoData app has specific design requirements for the record view.

#### Relevant ticket(s):

* [GDT-131](https://mitlibraries.atlassian.net/browse/GDT-131)
* [GDT-130](https://mitlibraries.atlassian.net/browse/GDT-130)

This adds a `record_geo` partial that includes metadata specific to GDT records, such as locations and a metadata download link. It also changes where the `access_button` partial is displayed depending on screen width, and it adds a 'back' button to the top of the record if `url_for(:back)` resembles a search results page.

* Some of the changes introduced in GDT-130  have been overwritten by this commit, as certain styles and markup are shared by both views. Specifically, the `geo_data_info` and `_authors` partials have been moved to the shared directory, and a new `shared` SCSS partial contains styles used in `result` and `record.
* Several helper methods have been added to parse the metadata needed for the record view.

#### Developer

#### Code Reviewer

NO

NO

[GDT-131]: https://mitlibraries.atlassian.net/browse/GDT-131?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[GDT-130]: https://mitlibraries.atlassian.net/browse/GDT-130?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/helpers/record_helper.rb
```diff
@@ -98,6 +98,48 @@ def access_type(metadata)
...
+    return if dates.blank? || dates&.none? { |date| date['kind'] == 'Issued' }
+
+    # If there is more than one date issued, take the first one.
+    dates.select { |date| date['kind'] == 'Issued' }&.uniq&.first['value']
+  end
+
+  def coverage_date(dates)
+    return if dates.blank? || dates&.none? { |date| date['kind'] == 'Coverage' }
+
+    # If there is more than one coverage date, take the first one.
```
JPrevost, line 111: I'm not sure this is the behavior we want. I feel we should display multiple dates in this situation, either as comma separate or as a bulleted list.

app/helpers/record_helper.rb
```diff
@@ -98,6 +98,48 @@ def access_type(metadata)
     access_right.first['description']
   end
 
+  def issued_date(dates)
+    return if dates.blank? || dates&.none? { |date| date['kind'] == 'Issued' }
+
+    # If there is more than one date issued, take the first one.
```
JPrevost, line 104: I'm not sure this is the behavior we want. I feel we should display multiple dates in this situation, either as comma separate or as a bulleted list.

app/views/record/_back_button.html.erb
```diff
@@ -0,0 +1,3 @@
+<% if url_for(:back)&.starts_with?([root_url, 'results'].join('')) %>
```
JPrevost, line 1: Nice.

app/views/shared/_geo_data_info.html.erb
```diff
@@ -0,0 +1,14 @@
+<ul class="list-inline">
+  <li><%= metadata['contentType']&.each { |type| type['value'] }&.join(' ; ') %></li>
```
JPrevost, line 2: I noted this in slack, but I think the css rules need to be updated for these `<li>` tags to include something like:

`vertical-align: top;`

to address weird alignment issues.

JPrevost (commented):
Unless we've heard otherwise that we want to silently drop multiple dates, I think we should update the logic to display the multiple values.

I have a suggestion on how to handle the weird alignment issues we can see in Safari we should either include in this or open as a ticket in the backlog so it doesn't keep popping up.

app/helpers/record_helper.rb
jazairi: What if I told you this is common?

JPrevost: Neat. I think this is a good example we may need to bring to DataEng and/or UX to decide if we want to change the data mappings/better understand the data mappings/better understand what we want to do on the UI side.

For now, if you want to build in duplicate detection logic in the UI I could get behind it. I'm reluctant to just assume all dates are duplicate (without more insight from DataEng as to how these mappings are created at least).

jazairi: Yeah, I generally don't like putting that kind of logic in UI, but it seems like a reasonable solution in this case.

JPrevost: I feel like it's still worth a convo between EngX/DataEng at some point to help us better understand all the assumptions about dates. But for now this looks good to me.

JPrevost (approved):

PR #140, opened 2024-03-22
Add skip links and refactor filter sidebar for accessibility

jazairi:

This application has many focusable elements, and it may be helpful
to users of assistive technology to provide skip links to different
sections.

#### Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/GDT-161

This pulls in the `application` layout from the theme gem and adds to it the following skip links:

* All views: nav, search form
* Results: results list, filter sidebar
* Record: record metadata

It also makes the filter sidebar more semantically meaningful. As currently written, VoiceOver does not parse the filter menus as disclosure triangles, and thus tabs through each available filter in a menu even while it is collapsed.. The menus are now written as `details` elements, which VoiceOver interprets correctly.

In the process of refactoring the filter sidebar, I also made a small change to the menu toggler. This had been assigned to an `onclick` attribute, but it is now more appropriately an event listener, and moved to the same file as the logic for the mobile filter toggler.

The main side effect is the added technical debt of maintaining a separate version of `application.html.erb`. This shouldn't be especially onerous, as that file is rarely updated. Additional side effects are noted in each commit message.

##### Accessibility

##### New ENV

- [x] ENV has not changed.

Note that this PR includes three commits: one to pull in the upstream layout, one to add the skip links, and one to refactor the filter sidebar. It will likely be easier to review by looking at each commit individually, from oldest to newest.

To verify the skip links, ensure that tab/screen reader navigation focuses on the following links before any other element, and they link to the correct section:
* Landing page: main menu (nav), search form (beginning with basic search)
* Results page: main menu, search form, search results, search filters
* Record page: main menu, metadata

To verify the filter refactor, ensure that tab/screen reader navigation focuses only on the filter menu label if it is collapsed. If the menu is open, it should focus first on the label, then on each available filter. Additionally, screen readers should read "apply filter" before available filters, and "remove filter" before applied filters.

You should also confirm that filter menus can be opened and closed, and that filters can be applied and removed, in both the mobile and desktop views.

##### Code

added technical debt.

##### Documentation

##### Testing


app/views/layouts/application.html.erb
```diff
@@ -0,0 +1,49 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <%= render partial: "layouts/head" %>
+  </head>
+  <body class="app-theme">
```
JPrevost, line 6: How would you feel about moving skip links in the theme to a partial so we can override just that partial in apps that need multiple?

app/views/layouts/application.html.erb
```diff
@@ -0,0 +1,49 @@
...
+    <a class="skip sr sr-focusable" href="#basic-search">Skip to search form</a>
+    <% if current_page?(results_path) %>
+      <a class="skip sr sr-focusable" href="#results">Skip to results</a>
+      <% if @filters.present? %>
+        <a class="skip sr sr-focusable" href="#filters">Skip to search filters</a>
+      <% end %>
+    <% end %>
+    <% if controller.controller_name == 'record' && controller.action_name == 'view' %>
+      <a class="skip sr sr-focusable" href="#full-record">Skip to metadata</a>
+    <% end %>
```
JPrevost, line 17: Are we confident we aren't making things worse?

https://webaim.org/techniques/skipnav/#multiple

JPrevost (commented):
I'll confirm the functionality shortly and work towards approving this.

I do have reservations about this feature and may want to discuss it further (not sure who would need to be involved quite yet) as I'm nervous we may be adding additional skip links when the actual problem is potentially how we present the content to begin with.

app/views/layouts/application.html.erb
jazairi: Can we ever be confident we aren't making things worse? 🙂 

That WebAim article is actually what brought me to the conclusion that multiple skip links are warranted for TIMDEX UI. Skip links to the results and filtersare necessary at minimum, I think, because tabbing through the form can be a really onerous process if geo search or advanced search (or, worse, both) are enabled.

I am much less sure about the skips to nav and to full record metadata, since both of those are near the top of the tab order (and 'skip to meadata' could just be 'skip to main content'). If you feel similarly, then I think this would be a good opportunity to follow up with Darcy for further clarification on why they're needed.

app/views/layouts/application.html.erb
jazairi: Great suggestion. I have a hard time imagining a need to override anything else in this layout.

app/views/layouts/application.html.erb
JPrevost: Thanks for your added perspective on how we got to this PR.

I still need to dig in a bit to better understand what options are doing here as so far I've just looked at code and didn't really get a feel for what was happening.

app/views/layouts/application.html.erb
JPrevost: We can open a ticket to add that feature to the theme gem and update this app to use it once available if that makes sense for working towards wrapping these UI features this sprint.

jazairi: I think I'll just update the theme gem as part of this ticket. It shouldn't take long. In the meantime, we can consider whether this is the right approach (i.e., if we want this many skip links, and this set of them).

app/views/layouts/application.html.erb
```diff
@@ -0,0 +1,49 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <%= render partial: "layouts/head" %>
+  </head>
+  <body class="app-theme">
+    <a class="skip sr sr-focusable" href="#main-menu">Skip to main menu</a>
```
JPrevost, line 7: Our skip links need `id=skip` to work... which we can't do because that would lead to multiple conflicting `ids`.

This discussion started on Slack, but not to the point of realizing (or at least I didn't realize at the time) that supporting multiple skip links will introduce a breaking change if we move from `id=skip` to `class=skip` at the theme gem level. We likely want to deprecate `id=skip` and move to to a new class name, such as `class=skiplink` to allow us to slowly transition all of our apps to the new behavior. All of the rails apps will get the new markup at the same time as they get the new CSS, but I believe omeka directly pulls the style guide CSS so we should take care to not break things when adjusting this feature at our style guide level.

We also noted that our current skip link CSS is broken in Safari. We should resolve that when moving from `id` to `class` to support this new feature. If you'd prefer to add timdex-ui css overrides for these changes and open a ticket to move them to the style guide/theme gem once we work out any kinks in this app that would be fine with me.

JPrevost (changes requested):
We need to at minimum address the id vs class for skip links. Either as timdex-ui overrides (with a ticket to move them to the theme/style guide) or directly in the theme/style guide.

We also need to resolve the bug with Safari not tabbing to our skip links. That could handled by adding an override style to timdex-ui and/or opening a ticket to resolve it at the theme gem/style guide level.

I believe you have already confirmed you will add a new feature to the theme gem to support overriding skip links rather than forking the template.

jazairi:
The latest commit updates the theme gem in order to move the changes to a skip links partial ([introduced here](https://github.com/MITLibraries/mitlibraries-theme/pull/60) and released in v1.4). For now, I addressed the Safari bug in TIMDEX UI only, but I opened [ENGX-260](https://mitlibraries.atlassian.net/browse/ENGX-260) to fix it in the style guide.

During our team meeting today, we discussed the possibility of wrapping multiple skip links in another tag, so we can continue to use the `#skip` selector rather than introducing a new class. I tried a few different iterations of this, and each of them caused the skip links to be invisible. I do think an unordered list would be more semantically meaningful here, but in the spirit of red-green-refactor, I suppose a series of `a` tags is good enough for now.

jazairi:
Before I merge, I will also move the addition of the new partial into its own commit, as is our practice, so it's clearer what the changes are..

JPrevost (approved):
Safari is still wonky, but I'd be okay with handling that as a separate ticket as it may require us a bit of time to work through the best way to handle it as the behavior is just weird.

jazairi:
Yeah, the Safari situation is looking increasingly rough. Adding the `:focus-visible` selector doesn't fix it, so we'll probably need a polyfill. I updated ENGX-260 with the article you found on the subject.

PR #148, opened 2024-03-28
Address UX feedback on GDT full record view

jazairi:

The 'more information' section of a GDT record should contain
all available fields, not just a selection. This is required
for initial usability testing, so stakeholders can decide
what fields they want to include.

#### Relevant ticket(s):

* [GDT-131](https://mitlibraries.atlassian.net/browse/GDT-131)

This adds all the fields that I suspect will appear in a GIS
record. It intentionally omits some, like holdings and
callNumbers, that I suspect will not. The display is intentionally
messy at this point, as not all of these fields will be in the
final product.

* I left in some convenience methods (and corresponding tests) that
I suspect will be useful at some point, even though they will not
be used in the current testing phase. I can remove these if their
presence is more confusing than helpful.
* Two new fields have been added to the record query, and cassettes
have been regenerated (in a separate commit).

##### Accessibility

##### New ENV

- [x] ENV has not changed.

* The query change/cassette regenerations are in a separate commit.
* As noted in the side effects, some convenience methods have been retained, and corresponding tests skipped. I wasn't sure if I wanted to keep these, but I have a hunch they will be useful at some point.
* Some record controller tests were skipped as they are currently failing due to an in-process data reload.
* I can't say with certainty that these are *all* the possible fields that may show up in a GIS record, but they're the ones that seem most likely to. It felt unreasonable to conditionally include every field returned by the record query, when many of them will probably always be nil. However, if the reviewer has more insight into this than I do, please let me know!

##### Code

added technical debt.

##### Documentation

##### Testing

[GDT-131]: https://mitlibraries.atlassian.net/browse/GDT-131?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/helpers/record_helper.rb
```diff
@@ -98,17 +98,23 @@ def access_type(metadata)
...
+  # for testing.
   def coverage_dates(dates)
     return_relevant_dates(dates, 'Coverage')
   end
 
+  # This method is likely to change post-MVP. Stakeholders need to see all available data for testing.
   def more_info?(metadata)
-    if issued_dates(metadata['dates']) || coverage_dates(metadata['dates']) || places(metadata['locations']) ||
-       metadata['provider']
+    if metadata['citation'] || metadata['dates'] || metadata['identifiers'] || metadata['languages'] ||
```
JPrevost, line 115: When we come back to this, let's see if we can find a way to refactor this so it isn't a super long conditional.

I'm okay with going with this for now, but as your comment notes we'll come back to it let's rethink how to approach it at that time.

test/controllers/record_controller_test.rb
```diff
@@ -55,6 +55,7 @@ def setup
     end
 
     test 'access button displays for freely accessible data' do
+      skip("DataEng has changed access type labels and is currently reloading dev1 as of 3/28 at 12:20pm.")
```
JPrevost, line 58: I think the data is reloaded. Maybe regen the cassettes and remove these skips?

JPrevost (changes requested):
I think we should regen the cassettes and unskip the tests before merge.

Non-blocking: I would slightly prefer we design the More Info section to include only what has been explicitly requested. If people what to see _what else_ is available a separate section that dumps all the data feels better to me. That said, if this is what feels best to you based on what you are hearing the stakeholders are looking for I can get over it.

I'm curious if we might approach the "more info" differently by providing a well thought out layout for everything we _think_ should be there and then just dumping the rest in a "maybe you care about this" section. I think TIMDEX UI already did that (or did at some point).

test/controllers/record_controller_test.rb
jazairi: @JPrevost The tests will still fail. We need to change the logic that determines how the access buttons are rendered, which I don't think belongs in this PR (it's ticketed [here](https://mitlibraries.atlassian.net/browse/GDT-252)).

jazairi:
@JPrevost Yeah, I agree that `more_info` should be refactored. It may also become less complicated once user testing is completed (i.e., if stakeholders decide they only want a few fields, then we don't need to worry about dumping everything in there).

JPrevost (approved):

PR #158, opened 2024-04-01
Render filter value names in non-GDT contexts

jazairi:

When introducing the helper method to rename GDT source filter values, we inadvertently removed all filter values for non-GDT apps. This is hilarious, but not very usable.

#### Relevant ticket(s):

N/A

This adds to the filter partial an `else` clause that renders the term key if GDT is not enabled.

Users will know what filters they are applying.

##### Accessibility

##### New ENV

- [x] ENV has not changed.

Successful application and removal (and visibility) of filters should be confirmed with and without the GDT feature flag.

##### Code

added technical debt.

##### Documentation

##### Testing

- [x] There are appropriate tests covering any new functionality.

### DIFF
--- a/app/views/search/_filter.html.erb
+++ b/app/views/search/_filter.html.erb
@@ -15,9 +15,11 @@
           <a href="<%= results_path(add_filter(@enhanced_query, category, term['key'])) %>">
           <span class="sr">Apply filter:</span>
         <% end %>
-          <% if Flipflop.enabled?(:gdt) %>
-            <span class="name"><%= gdt_sources(term['key'], category) %></span>
-          <% end %>
+        <% if Flipflop.enabled?(:gdt) %>
+          <span class="name"><%= gdt_sources(term['key'], category) %></span>
+        <% else %>
+          <span class="name"><%= term['key'] %></span>
+        <% end %>
           <span class="count"><%= term['docCount'] %> <span class="sr">records</span></span>
         </a>
       </li>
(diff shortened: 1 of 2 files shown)


app/views/search/_filter.html.erb
```diff
@@ -15,9 +15,11 @@
           <a href="<%= results_path(add_filter(@enhanced_query, category, term['key'])) %>">
           <span class="sr">Apply filter:</span>
         <% end %>
-          <% if Flipflop.enabled?(:gdt) %>
-            <span class="name"><%= gdt_sources(term['key'], category) %></span>
-          <% end %>
+        <% if Flipflop.enabled?(:gdt) %>
+          <span class="name"><%= gdt_sources(term['key'], category) %></span>
+        <% else %>
```
JPrevost, line 20: 🤣 🙈 

Yeah, that indeed is what we are seeing :) 

I'm now sort of wondering if `gdt_sources` is a misleading method name as I sort of expect that to only affect the `source` filter and not all filters. I feel that not strongly enough to suggest changing it as part of this fixup.

JPrevost (approved):

app/views/search/_filter.html.erb
jazairi: I struggled with that, too. We decided to refactor `gdt_sources` to return the value as-is if it's not a source as a safety measure, but it may be more readable to remove that from the helper method and go back to something like `gdt_sources(term['key'], category) || term['key']`? 🤷

PR #162, opened 2024-04-10
Display filters in search summary in order of application

jazairi:

The search summary currently displays filters in alphabetical order by key. This is confusing, as one would expect them to show up in the order they were applied.

#### Relevant ticket(s):

* [GDT-224](https://mitlibraries.atlassian.net/browse/GDT-224)

Many (if not all) Rails helpers pertaining to URLs (`to_query`, `to_params`, routing helpers, etc.) sort their inputs lexicographically. This means that we cannot use any of these methods, or the `params` object, to pass query params into the Enhancer, as they will not remain ordered and will instead be alphabetized everywhere in the application.

This commit seeks to fix this problem by passing
`request.original_url` into the Enhancer and building filter URLs manually. It's messier than using the Rails convenience methods, particularly because we're now mostly working with strings rather than hashes, but it felt less complex than the other approach considered of storing filter order in a cookie.

Prior to this change, both the Enhancer and QueryBuilder iterated over arrays of supported params and added them to the query if they showed up in the `params` object. This is now reversed: the extraction methods in both models now iterate over the URL params and confirm that they exist in the list of supported params before adding them to the query. The reason for this change is to ensure that filters (and other params) are added to the query in the order in which they appear in the original URL, not the order in which they are listed in the array of supported params.

* Fixes a bug in the filter helper that prevents single-valued filters from being applied. Currently, the only single-valued filter we have is `literaryFormFilter`, which is not used anywhere, but it makes sense to support all fields that TIMDEX API provides.
* Updates several tests to accommodate the refactored Enhancer and QueryBuilder.
* Moves the constants that recur in the Enhancer and QueryBuilder into a new module, QueryElements.
* Moves all of the filter helper methods related to adding/removing filters next to each other for better readability.
* Regenerates several cassettes. (It's possible that not all of these needed to be regenerated; I erred on the side of more.)

##### Accessibility

##### New ENV

- [x] ENV has not changed.

In terms of confirming functionality: ensure that filters from _every_ category (including `literaryFormFilter`) can be added and removed successfully, and appear in the search summary panel in the order you add them.

In terms of the code: one thing I am particularly interested in is where this is adequately documented. Please let me know if (and where) you'd like to see additional explanation.

##### Code

added technical debt.

##### Documentation

##### Testing

[GDT-224]: https://mitlibraries.atlassian.net/browse/GDT-224?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/helpers/filter_helper.rb
```diff
@@ -12,48 +12,29 @@ def add_filter(query, filter, term)
...
-  end
-
-  def gdt_sources(value, category)
-    return value if category != :sourceFilter
-
-    return 'Non-MIT institutions' if value == 'opengeometadata gis resources'
-
-    return 'MIT' if value == 'mit gis resources'
-
-    value
```
jazairi, line 43: Note to reviewer: this is the result of reordering the methods in the filter helper, as noted in the side effects. (See also lines 70-123.)

app/helpers/filter_helper.rb
```diff
@@ -78,8 +59,66 @@ def applied_filters(query)
...
+    return 'Non-MIT institutions' if value == 'opengeometadata gis resources'
+
+    return 'MIT' if value == 'mit gis resources'
+
+    value
+  end
+
+  private
+
+  # This kind of manual URL construction is required because any core exntesions method involving params, including
```
JPrevost, line 97: Typo: `exntesions`

jazairi:
I noted this in JIRA, but there's a bug in this PR that still allows filters to be displayed out of order.

jazairi:
I noted this in JIRA, but there's a bug in this PR that still allows filters to be displayed out of order.

In the query object, multiple filters in the same category are assigned to the same key, since keys are unique in Ruby hashes.

This means that when we apply the filters in the example below (subject: transportation, then place: massachusetts, then subject: maps), the query is built like so:

```
{
  subjectsFilter: ['transportation', 'maps'],
  locationsFilter: ['massachusetts']
}
```

And when we iterate through that hash, the filters get inevitably grouped by key:

```
[{ subjectsFilter: 'transportation', subjectsFilter: 'maps', locationsFilter: 'massachusetts' }]
```

I think we need to rework our approach to this problem. It's likely a good use case for cookies, as we've discussed elsewhere.

PR #166, opened 2024-04-18
Add affordance to clear search terms

jazairi:

Now that we have several input fields in the search form, it would be helpful to have a button that can clear everything in the form.

#### Relevant ticket(s):

* [GDT-244](https://mitlibraries.atlassian.net/browse/GDT-244)

This adds a 'clear' button to the form, with a corresponding click handler that triggers a method to clear all inputs, except hidden ones.

This uses the `querySelectorAll()` method, which is [supported by all modern browsers](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelectorAll#browser_compatibility) but may not be supported by legacy browsers. This feels like an acceptable risk.

An alternative approach is `getElementsByTagName`. This would also select the hidden elements, which toggle geospatial and advanced searches. I've avoided this because it may lead to some unexpected behavior, but it also feels relatively low-risk.

##### Accessibility

##### New ENV

- [x] ENV has not changed.

We are not simply linking to the root route because we want to retain context about the current search.

##### Code

added technical debt.

##### Documentation

##### Testing

[GDT-244]: https://mitlibraries.atlassian.net/browse/GDT-244?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/javascript/search_form.js
+++ b/app/javascript/search_form.js
@@ -97,10 +97,17 @@ function enableGeodistance() {
   geodistance_label.innerText = 'Close distance search';
 };
 
+// Note that this skips hidden inputs, which are used to toggle advanced and geospatial searches.
+function clearInputs(event) {
+  [...search_form.querySelectorAll('input:not([type=hidden])')].forEach(function(field) {
+    field.value = '' ;
+  });
+};
 
 var advanced_field = document.getElementById('advanced-search-field');
 var advanced_label = document.getElementById('advanced-search-label');
 var advanced_toggle = document.getElementById('advanced-summary');
+var clear_button = document.getElementById('clear-button');
 var details_panel = document.getElementById('advanced-search-panel');
 var keyword_field = document.getElementById('basic-search-main');
 var geobox_field = document.getElementById('geobox-search-field');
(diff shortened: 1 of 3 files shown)


app/javascript/search_form.js
```diff
@@ -97,10 +97,17 @@ function enableGeodistance() {
   geodistance_label.innerText = 'Close distance search';
 };
 
+// Note that this skips hidden inputs, which are used to toggle advanced and geospatial searches.
+function clearInputs(event) {
```
JPrevost, line 101: Can we not use https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/reset because we don't want those hidden inputs cleared?

jazairi: `No,` `reset()` meets a different use case. It resets the form to its original values (i.e., what they were when the current page loaded). The requested feature is to clear all form values.

jazairi:
@JPrevost FYI, Darcy agreed to just make this a link to the root route, so I'm going to close this and open a new PR to that effect.

PR #189, opened 2024-05-10
Consolidate full record access links in sidebar

jazairi:

The metadata download link appears at the end of the full record with no heading. This makes it look like part of the last full record heading (usually 'Languages').

A related issue is that the access link in the sidebar has no heading. We have an `aria-label` that describes this section for screen reader users, but for sighted users, this section lacks context.

As part of their accessibility review, DAS has recommended moving the metadata download link to the same section as the access link.

#### Relevant ticket(s):

* [GDT-319](https://mitlibraries.atlassian.net/browse/GDT-319)

This moves the download link below the access link in the sidebar, and adds a heading to that section.

While working on this, I noticed that the fact panel markup needs some work to be semantically meaningful. I haven't touched that in this commit, but it will be worth revisiting when we return to search intent detection.

##### Accessibility

##### New ENV

- [x] ENV has not changed.

##### Code

added technical debt.

##### Documentation

##### Testing

[GDT-319]: https://mitlibraries.atlassian.net/browse/GDT-319?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/views/record/_access_buttons.html.erb
+++ b/app/views/record/_access_buttons.html.erb
@@ -1,6 +1,6 @@
 <% return if @record.blank? %>
 
-<div class="access-button-container <%= display %>">
+<div class="access-button-container">
   <% if access_type(@record) == 'no authentication required' %>
     <a class="btn button-primary access-button" href="<%= gis_access_link(@record) %>">Download geodata files</a>
   <% elsif access_type(@record) == 'MIT authentication required' %>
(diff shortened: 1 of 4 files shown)


app/views/record/_access_buttons.html.erb
```diff
@@ -13,4 +13,10 @@
       View <%= @record['provider'] %> record
     </a>
   <% end %>
+  <% if access_type(@record) != 'unknown: check with owning institution' &&
```
JPrevost, line 16: This is fine as it was like this before moving the location of where it shows up, but I'm curious if we actually need the check to make sure it is an MIT record or if we should just always display the download metadata link if it is present. Not blocking, but suspect we're being overly cautious here.

JPrevost (approved):

app/views/record/_access_buttons.html.erb
jazairi: Yeah, good point. I think this made more sense when that value was 'Not owned by MIT'. Now that the value itself doesn't make sense, it's probably best just to use `provider`.

JPrevost: I don't think we should check either. I think we should just display this link if it exists.

jazairi: Oh. You're probably right about that, too.

I'd like to revisit that when we (hopefully) talk about displaying geodata file download links for all records, post-MVP. For now I agree it's fine as-is.

PR #191, opened 2024-05-16
Redesign contributors partial

jazairi:

The authors partial lists all of the authors in a `p` tag, delimited
with a semicolon. An unordered list would be more semantically
meaningful.

UXWS has also suggested that we should include contributor types
in the list, to make clearer what their contribution to the work.

#### Relevant ticket(s):

* [GDT-284](https://mitlibraries.atlassian.net/browse/GDT-284)

This refactors the authors partial to be an unordered list and
renders the contributor type alongside the name value. It also renames
the authors partial to contributors, which more accurately describes
its use.

This does not touch the non-GDT record view, as it uses somewhat
different logic to render contributors. However, this might be
an opportunity to normalize, unless there's a valid reason why
the logic differs.

##### Accessibility

##### New ENV

- [x] ENV has not changed.

N/A

##### Code

added technical debt.

##### Documentation

##### Testing


app/views/search/_result_geo.html.erb
```diff
@@ -8,9 +8,7 @@
   </div>
 
   <% if result_geo['contributors'] %> 
-    <p class="authors truncate-list">
```
matt-bernhardt, line 11: What happens to the `truncate-list` class in this partial? Are we not using it anymore?

app/views/shared/_contributors.html.erb
```diff
@@ -0,0 +1,16 @@
+<%= return if contributors.blank? %>
+
+<% contribs_list = [] %>
+<% contributors&.each do |contributor| %>
+  <% contribs_list << contributor %>
+<% end %>
+
+<span class="sr">Contributors: </span>
+<ul class="list-inline contributors">
```
matt-bernhardt, line 9: I don't think there's anything to do here, but I was confused when I saw the `list-inline` class here but saw the review app still showing each element on its own line. Then I saw that the `li` elements in this context pick up an `inline-block` style, which prevents them from spanning lines - which explains the behavior.

Without this class we get a bulleted list, which doesn't seem like what we want - but I guess I'm just noting that "inline" here may be a bit of a misnomer.

matt-bernhardt (approved):
I've got one question, and one non-blocking comment. My question, which I'm probably happy to approve either way, is about the `truncate-list` class that seems like it might be getting lost in the shuffle. If that's intentional, then I'm okay with this merging. If that wasn't what we intended, then should the class name be preserved?

My other question is more of a note to future-us about what we might name classes like `list-inline`.

I acknowledge, though, that I might not have the needed context for a review here - so if Jeremy would be better, please ignore me. :-)

app/views/shared/_contributors.html.erb
jazairi: I can't speak to why the style guide uses `inline-block` instead of `inline`. If I had to guess, it's probably because it results in more predictable behavior, since width/height can be set. Either way, I take your point that `list-inline-block` would be more descriptive, and a good note to recall when we rebuild the style guide.

app/views/search/_result_geo.html.erb
jazairi: Ohh, good catch. We *should* still be using it, but I was writing this code during a conference and not paying very much attention. 🙂 Change forthcoming...

jazairi:
@matt-bernhardt I just pushed up a commit that hopefully addresses the truncation issue (i.e., the lack thereof). Ironically, this also required overriding the `inline-block` rule applied by `list-inline`. (More context on this in the commit message.)

app/views/shared/_contributors.html.erb
matt-bernhardt: Agreed on all fronts here...

matt-bernhardt (approved):
This looks good to me - I'm glad I wasn't being too picky in noticing the change to the class list.

Good catch, also, on the helper method - I'm apparently not close enough to the data to realize that this was needed.

jazairi:
I'm also not close enough to the data, unfortunately, which is why it took me a second look to determine that it was needed. 😅 I'm sure there are other oddities that we should catch, but ideally all of that should happen in the ETL pipeline (unless values like `mitauthor` are useful for reasons I can't think of).

JPrevost:
@jazairi fwiw `mitauthor` is probably more hopeful than useful in the data model and is very unlikely not populated for any non-experimental sources (aka I think it came from RDI).

jazairi:
@JPrevost I think you're right. What's weird is that `contributor` also provides an `mitAffiliated` boolean that serves the same purpose. Maybe we had `mitauthor` first and didn't go back to change it? 🤷

JPrevost:
@jazairi 🙈

jazairi:
Closing this without merging. After a round of QA, Darcy has decided this is not needed.

PR #192, opened 2024-05-20
Download links target _blank and append UI marker

JPrevost:

* We want to display a message in our restricted CDN auth flow that instructs a user to close the window/tab when the download is complete to return to their search if they came from TIMDEX UI.

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/GDT-317
* https://mitlibraries.atlassian.net/browse/GDT-304

* Adds `target=_blank` to download link tags
* appends `?timdex-ui=true` to download link URLs

* A bunch of tests reformatted automatically on save. The only relevant changes to the tests were on lines 246-247

##### Accessibility

##### New ENV

- [ ] ENV has not changed.

The download links for restricted MIT downloads should now pop into a new window/tab. The messaging for the CDN auth app has not been updated to inform the user to close the window to return to the search, but that will be added soon.

All download links (restricted and not) should get `?timdexui=true` appended.

Links to metadata or external sites should not be affected.

##### Code

added technical debt.

##### Documentation

##### Testing


app/helpers/record_helper.rb
```diff
@@ -86,7 +86,8 @@ def gis_access_link(metadata)
     if access_type(metadata) == 'unknown: check with owning institution'
       links.select { |link| link['kind'] == 'Website' }.first['url']
     else
-      links.select { |link| link['kind'] == 'Download' && link['text'] == 'Data' }.first['url']
+      url = links.select { |link| link['kind'] == 'Download' && link['text'] == 'Data' }.first['url']
+      "#{url}?timdexui=true"
```
matt-bernhardt, line 90: My only question is whether there's going to be any of these links that already have a querystring in them, which would result in something like

```
https://cdn.dev1.mitlibrary.net/geo/public/US_MA_P3STATIONS_2004.zip?foo=bar?timdexui=true
```

matt-bernhardt (commented):
This all looks good to me, and I've confirmed that the one relevant test makes sense with this change to the helper and the template. I do have a question about whether any of the URLs passing through this system already have querystrings - in which case I think more complex logic might be needed in the helper.

If that's not in our data (and isn't likely to be in the future), then I'm fine with all of this and will switch this to an approval.

matt-bernhardt (approved):
That looks good - I appreciate having those additional helper methods in place here, because that would have been a lot to parse without them.

:shipit:

PR #194, opened 2024-05-29
Improve markup for contributors

jazairi:

Authors are currently rendered within a `p` tag. This is not very semantically meaningful.

#### Relevant ticket(s):

* [GDT-284](https://mitlibraries.atlassian.net/browse/GDT-284)

This refactors the contributors section to a partial and renders contributors as an unordered list. It is the same changeset that was introduced in #191, but without displaying the contributor type, which UXWS has deemed unnecessary at this time.

None.

##### Accessibility

##### New ENV

- [x] ENV has not changed.

N/A

##### Code

added technical debt.

##### Documentation

##### Testing

[GDT-284]: https://mitlibraries.atlassian.net/browse/GDT-284?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/views/search/_result_geo.html.erb
+++ b/app/views/search/_result_geo.html.erb
@@ -8,9 +8,10 @@
   </div>
 
   <% if result_geo['contributors'] %> 
-    <p class="authors truncate-list">
-      <%= render partial: 'shared/authors', locals: { contributors: result_geo['contributors'] } %>
-    </p>
+    <span class="sr">Contributors: </span>
+    <ul class="list-inline truncate-list contributors">
+      <%= render partial: 'shared/contributors', locals: { contributors: result_geo['contributors'] } %>
+    </ul>
   <% end %>
 
   <% if result_geo['summary'] %>
--- a/app/views/shared/_contributors.html.erb
+++ b/app/views/shared/_contributors.html.erb
@@ -1,0 +1,5 @@
+<%= return if contributors.blank? %>
+
+<% contributors.uniq do |contributor| %>
+  <li><%= link_to contributor['value'], results_path({ advanced: true, contributors: contributor['value'] }) %></li>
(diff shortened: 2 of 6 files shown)


app/views/shared/_contributors.html.erb
```diff
@@ -0,0 +1,5 @@
+<%= return if contributors.blank? %>
+
+<% contributors.uniq.map do |contributor| %>
```
JPrevost, line 3: Is `.map` doing something here I'm not understanding?

app/views/search/_result_geo.html.erb
```diff
@@ -8,9 +8,10 @@
   </div>
 
   <% if result_geo['contributors'] %> 
-    <p class="authors truncate-list">
-      <%= render partial: 'shared/authors', locals: { contributors: result_geo['contributors'] } %>
-    </p>
+    <span class="sr">Contributors: </span>
```
JPrevost, line 11: I think it would be a bit of a refactor to do this, so I don't think we should in this change... but if we changed `result_geo` to just be `result` wherever it came from we'd be able to remove some duplication and move the entire contributors logic into the shared partial instead of having geo/normal variants that are identical.

JPrevost (changes requested):
Can you look into whether the `map` I noted is doing anything meaningful? I ran through a few searches with it as-is and with it removed and don't see any difference but I'm also pretty early into my coffee...

app/views/shared/_contributors.html.erb
jazairi: Yeah, good call. It's probably a remnant of a previous iteration.

PR #195, opened 2024-05-29
Fix form required attributes redundancy and keyword input description

jazairi:

DAS has informed us that it is bad practice to use both `required` and `aria-required`. They also suggested that we add an `aria-describedby` to the keyword input that links to the `p` tag below the site title.

#### Relevant ticket(s):

* [GDT-327](https://mitlibraries.atlassian.net/browse/GDT-327)

This removes the redundant `aria-required` attributes and makes the suggested labeling change for the keyword input.

The labeling piece only affects GDT, as non-GDT apps don't currently use the `site_title` partial. This feels acceptable, because the keyword input is more clearly labeled in other TIMDEX UI applications.

##### Accessibility

##### New ENV

- [x] ENV has not changed.

Confirm that required fields are still required (i.e., built-in form validations fire when they are empty).

##### Code

added technical debt.

##### Documentation

##### Testing

[GDT-327]: https://mitlibraries.atlassian.net/browse/GDT-327?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/views/search/_form.html.erb
+++ b/app/views/search/_form.html.erb
@@ -30,7 +30,8 @@ end
     <input id="basic-search-main" type="search"
            class="field field-text basic-search-input <%= "required" if search_required %>" name="q"
            title="Keyword anywhere" placeholder="Enter your search"
-           value="<%= params[:q] %>" <%= 'required="required" aria-required="true"' if search_required %>>
+           value="<%= params[:q] %>" <%= 'required' if search_required %>
+           <%= 'aria-describedby=site-desc' if Flipflop.enabled?(:gdt) %>>
 
     <% if Flipflop.enabled?(:gdt) %>
       <details id="geobox-search-panel" <%= "open" if params[:geobox] == "true" %>>
@@ -47,7 +48,7 @@ end
               <input type="number" step="0.000001" min="-180.0" max="180.0"
                      class="field field-text <%= "required" if geobox_required %>" 
                      id="geobox-minLongitude" name="geoboxMinLongitude" value="<%= params[:geoboxMinLongitude] %>"
-                     <%= 'required="required" aria-required="true"' if geobox_required %>
+                     <%= 'required' if geobox_required %>
                      aria-describedby="minLong-desc">
               <span class="geo-desc" id="minLong-desc">
                 A decimal between -180.0 and 180.0 (Ex: -73.507239)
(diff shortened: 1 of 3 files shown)


jazairi:
CodeClimate issues are valid and noted, but not new. I think they showed up again because I touched the offending JS, so it analyzed it again instead of ignoring it.

jazairi:
Rich has confirmed that this review app fixes the last of the a11y issues. Opening for review.

app/views/search/_form.html.erb
```diff
@@ -30,7 +30,8 @@ end
     <input id="basic-search-main" type="search"
            class="field field-text basic-search-input <%= "required" if search_required %>" name="q"
            title="Keyword anywhere" placeholder="Enter your search"
-           value="<%= params[:q] %>" <%= 'required="required" aria-required="true"' if search_required %>>
+           value="<%= params[:q] %>" <%= 'required' if search_required %>
+           <%= 'aria-describedby=site-desc' if Flipflop.enabled?(:gdt) %>>
```
JPrevost, line 34: Can you help me understand why this is only being used for GDT enabled sites? I noted you called this out as unnecessary for non-GDT in the commit message but I don't understand why. I think I see the differences, but it's unclear why it was only an issue in GDT.

jazairi: The relative clarity of the non-GDT form is the tiebreaker for me. It's just a keyword input with some advanced fields, so I suspect it will be clearer if a screen reader user goes from an`h2` with the text of 'Search the MIT Libraries' to a 'Keyword anywhere' input.

To be transparent, though, this is an assumption I've made. I did not ask Rich to test search.libraries.mit.edu, and it's possible he would have the same feedback on that application. If that's a concern, I'd be happy add some copy (with Darcy's input) to describe the keyword input for non-GDT apps.

JPrevost: I'm fine with this solution for now as GDT is the only prod app. I'm waffly on whether we should pre-address this for non-GDT apps or remember/expect to come back to it when we prep non-GDT apps for prod. I'll go ahead and approve this.

JPrevost (approved):

PR #198, opened 2024-06-07
Fix iOS text wrap bug

jazairi:

SwiftUI/iOS has some unusual behavior for list markers. It adds line height to them, such that when we increase the font size for a marker, the line height is dramatically increased. This is particularly visible when the text next to the marker wraps, which is common for TIMDEX UI record titles.

#### Relevant ticket(s):

* [GDT-246](https://mitlibraries.atlassian.net/browse/GDT-246)

This implements a CSS workaround. First, we reset the line height to 0 on `.result`, enforcing no line height for markers. Then, we add a div to wrap the rest of the result (`.result-content`), with a rule that sets the line height to 1.2.

I don't forsee any, but it's a pretty ugly workaround to accommodate a minor bug in a weird OS.

##### Accessibility

##### New ENV

- [x] ENV has not changed.

This will get UXWS review, but if the code reviewer would like to do QA as well, it would be useful to confirm the changes on a few browsers on desktop and on an iOS device. 

##### Code

added technical debt.

##### Documentation

##### Testing

[GDT-246]: https://mitlibraries.atlassian.net/browse/GDT-246?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/assets/stylesheets/partials/_results.scss
+++ b/app/assets/stylesheets/partials/_results.scss
@@ -44,16 +44,11 @@
   padding: 2rem;
   border-top: 1px solid $brand-primary;
   font-size: $fs-small;
+  line-height: 0; /* line height needs to be reset so iOS doesn't act weird */
 
   &::marker {
     font-size: 2.5rem;
-    line-height: 1.1;
     font-weight: $fw-bold;
-  }
-
-  .record-title {
-    font-size: 2.5rem;
-    line-height: 1.1;
   }
 
   &:hover,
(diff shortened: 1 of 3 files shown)


JPrevost:
Interesting... I wonder if our "CSS reset" settings might be expanded to cover things like this without breaking things badly:
https://github.com/MITLibraries/mitlib-style/blob/master/_assets/sass/global/_unsets.scss

jazairi:
> Interesting... I wonder if our "CSS reset" settings might be expanded to cover things like this without breaking things badly: https://github.com/MITLibraries/mitlib-style/blob/master/_assets/sass/global/_unsets.scss

It would take some creativity. We might be able to do something like this:

```css
li::before {
  line-height: 0
}
li::after {
  line-height: 1 /* or whatever */
}
```

JPrevost (commented):
I'm not sure I agree this is an improvement. I'll bring a conversation to slack tomorrow to make sure I understand what is changing and why and share what I am seeing to ensure we're seeing the same thing before we confirm if this addresses the problem

jazairi:
@JPrevost Could you confirm that the two recently pushed commits address the issue? I'd moved the padding down a level in the markup so the hover/focus rule would include it, but in retrospect I should've just moved the hover/focus rule up a level.

JPrevost (approved):

jazairi:
Approved by UXWS this morning. Merging now.

PR #219, opened 2025-08-07
Add ADR for surfacing Primo CDI records in results

jazairi:

The proposed unified search interface would display records from Primo CDI (via Primo Search API) and Alma (via TIMDEX API) in the same results list. TIMDEX UI does not currently have a means to combine results from multiple APIs in this way.

#### Relevant ticket(s):

N/A

This adds an ADR that outlines a proposed solution to this problem, by introducing a search orchestration layer that will handle API calls and results normalization.

There are additional decisions to be made around the architecture of the search orchestrator, such as how to manage relevance normalization. These decisions are noted in the ADR and will be explored in future ADRs.

##### Accessibility

- [x] This PR contains no changes to the view layer.

##### New ENV

- [x] ENV has not changed.

- [x] UXWS/stakeholder review is not needed.

There's some hand-waving regarding implementation details. From my perspective, it seemed adequate to delay these decisions until we start working on the orchestrator, but I would be happy to modify the ADR to include more detail if that would be useful.

##### Code

added technical debt.

##### Documentation

##### Testing

### DIFF
--- a/docs/architecture-decisions/0003-surface-primo-cdi-records-in-results.md
+++ b/docs/architecture-decisions/0003-surface-primo-cdi-records-in-results.md
@@ -13,22 +13,46 @@ both Primo Central Discovery Index (CDI) and Alma (via TIMDEX), replacing the cu
 In Bento, Alma and CDI results are displayed in separate boxes. The unified interface would
 interleave CDI and TIMDEX records in the same results list.
 
+## Options considered
+
+### Harvest Primo CDI data
+
 We considered adding a new Primo harvester to our ETL architecture to ingest CDI data into TIMDEX
-API. However, this approach is not feasible for many reasons:
+API. This would allow us to normalize CDI records as we do with other TIMDEX sources. Querying a
+single API for Alma and CDI records would facilitate a single-stream view as desired in the unified
+UI. Interleaving would no longer be necessary, as all records would be stored in OpenSearch.
+
+The harvester model would value beyond the scope of the TIMDEX UI redesign. By storing CDI records
+in TIMDEX API, we could facilitate computational access to a massive corpus of data.
+
+Unfortunately, this approach is not feasible for many reasons:
 
 - **Cost**: CDI contains over 5 billion records. Harvesting and storing these records would be impractical and expensive, both in terms of financial and compute resources.
 - **Performance**: Expanding TIMDEX API at such a scale is likely to dramatically reduce the efficiency of our OpenSearch index.
 - **Data availability**: Because Primo does not expose CDI records in OAI-PMH, we would need to harvest using the Primo Search API, making the process needlessly complex and perhaps impossible.
 - **Licensing**: Harvesting CDI records for TIMDEX likely has licensing implications. Ex Libris seems to discourage the practice, as Primo does not provide OAI-PMH support, and the Search API caps records per request at 5,000 via the [`offset` parameter](https://developers.exlibrisgroup.com/primo/apis/docs/primoSearch/R0VUIC9wcmltby92MS9zZWFyY2g=/#output:~:text=Note%3A%20The%20Primo%20search%20API%20has%20a%20hardcoded%20offset%20limitation%20parameter%20of%205000.).
 
+### Display separate result streams in tabbed views
+
+This option would essentially be a different take on the Bento design. On the results page, a user
+could tab between Alma results (labeled 'Books', 'MIT Catalog', etc.) and CDI results ('Articles').
+
+While arguably an improvement on Bento, this design does not deliver the combined Alma/CDI results
+view as envisioned in the unified UI. 
+
+### Implement external search orchestrator
+
+In this approach, we would surface CDI records in TIMDEX UI by querying the Primo Search API
+directly at runtime and interleaving results with TIMDEX API results in the unified search
+interface.
+
+To achieve this, we would implement a search orchestrator that receives a query from TIMDEX UI and
+dispatches it in parallel to TIMDEX API and Primo Search API. The orchestrator would normalize and
+interleave the results before returning them to the UI. This would allow us to display Alma and
+CDI results in the same results list, without the feasibility concerns inherent in ingesting CDI
+records into TIMDEX API.
+
 ## Decision
-
-We will surface CDI results in TIMDEX UI by querying the Primo Search API directly at runtime and
-interleaving results with TIMDEX API results in the unified search interface.
-
-To achieve this, we will implement a search orchestrator that receives a query from TIMDEX UI and
-dispatches it in parallel to TIMDEX API and Primo Search API. The orchestrator will normalize and
-interleave the results before returning them to the UI.
 
 This approach aligns with the unified search strategy's goal to display all known results from
 CDI and TIMDEX in the same interface. It also enables us to add the desired intelligent user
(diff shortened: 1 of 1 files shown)


matt-bernhardt (commented):
I'm not in a position right now to indicate a vote or approval/change request on this text - I think we're in a discussion phase prior to concluding anything, so that's what I'm focusing on for now...

* While I agree with the prospect of building an orchestrator and proceeding with a simple interleaving logic for merging the two sets of results, I think formally I'd like this document to describe multiple options prior to stating the decision we're going with. The one clear alternative I see right now would be to harvest CDI records into TIMDEX, although I think others are implicitly referenced (continuing to display the two results separately without merging, building an external orchestrator in Python, building an internal orchestrator in Rails).

---
* The sequence diagram feels like it could be cleaner in its placement of the "User submits search query" phase - maybe that comes in from a User participant to the frontend, which then delegates the search to both TACOS and the orchestrator?

I'm also not sure whether any of the interactions in the diagram are blocking each other - it feels like the basic phases are meant to be:

1. User submits search query to TIMDEX UI

2. TIMDEX UI starts processing, doing internal things

3. TIMDEX UI issues the search to three targets: TACOS, TIMDEX API, and the CDI (these later two may be managed by the Search Orchestrator, while some other model in the application handled TACOS?)

4. As responses are received from the three external systems, some (TACOS) are displayed immediately, in parallel to the other streams. The Search Orchestrator, however, waits to respond until it has received responses from both TIMDEX API and the CDI. It interleaves those responses on its own before returning them for rendering (which is also a difference to TACOS, whose response is not modified in any way)

5. As the responses are processed in step 4, they are added to the DOM and shown to the user.

I think this is pretty much the same sequence you're proposing - so I think I'm ultimately wanting to have a diagram that makes clear what the dependencies are, and what might be happening in parallel. This might be my misunderstanding how to read a sequence diagram, however.

---
* I'm not sure whether the TIMDEX UI is the most appropriate place to consider how to facilitate computation access to CDI records. While it is definitely a concern for the Libraries, I think it makes sense to discuss that use case elsewhere? The choice to build an orchestrator or adopt a separate approach feels completely separate from that use case (although I might be missing something, so I'm not yet at a point of blocking it being here)

---
I think those are the thoughts which are most prominently occurring to me at the moment. I may have other comments in the next day or so, prior to our discussing this as a team.

Thanks for putting this together! I appreciate the work it took to assemble.

docs/architecture-decisions/0003-surface-primo-cdi-records-in-results.md
```diff
@@ -0,0 +1,146 @@
...
+Date: 2025-08-07
+
+## Status
+
+Accepted
+
+## Context
+
+The Libraries' unified search strategy calls for a discovery interface that surfaces results from
+both Primo Central Discovery Index (CDI) and Alma (via TIMDEX), replacing the current [Bento UI](https://github.com/MITLibraries/bento).
```
JPrevost, line 12: We call it Primo CDI, but it's actually  [Ex Libris Central Discovery Index (CDI)](https://knowledge.exlibrisgroup.com/Primo/Content_Corner/Central_Discovery_Index/Documentation_and_Training/Documentation_and_Training_(English)/CDI_-_The_Central_Discovery_Index/010An_Overview_of_the_Ex_Libris_Central_Discovery_Index_(CDI))

docs/architecture-decisions/0003-surface-primo-cdi-records-in-results.md
```diff
@@ -0,0 +1,146 @@
...
+
+To achieve this, we would implement a search orchestrator that receives a query from TIMDEX UI and
+dispatches it in parallel to TIMDEX API and Primo Search API. The orchestrator would normalize and
+interleave the results before returning them to the UI. This would allow us to display Alma and
+CDI results in the same results list, without the feasibility concerns inherent in ingesting CDI
+records into TIMDEX API.
+
+## Decision
+
+This approach aligns with the unified search strategy's goal to display all known results from
```
JPrevost, line 57: I don't believe you stated which approach was decided.

docs/architecture-decisions/0003-surface-primo-cdi-records-in-results.md
```diff
@@ -0,0 +1,146 @@
...
+### Future Considerations
+
+Usability testing and analytics will inform how we refine this feature. Depending on how users
+interact with the single-stream UI, we may need visual clarification of each record's source API, or 
+separate tabs for TIMDEX and Primo records.
+
+Relevance normalization is a critical issue. We can begin with rank-based interleaving, but we
+should not assume this to be a long-term solution.
+
+We should connect with the MIT research community to determine whether computational access to CDI
```
JPrevost, line 143: I'm fairly certain there is no legal way to obtain these without violating our contract to access the records in the first place. It would be better for us to look at what our user needs are more broadly and understand whether there is a CDI-alternative that would better meet their needs (either as part of TIMDEX or as a recommendation to use instead of TIMDEX)

docs/architecture-decisions/0003-surface-primo-cdi-records-in-results.md
```diff
@@ -0,0 +1,146 @@
...
+
+### Pros
+
+- Avoids duplicating CDI data or violating licensing terms.
+- Enables real-time access to CDI content via Primo Search API.
+- Supports the unified search vision without overloading TIMDEX API.
+
+### Cons
+
+- Requires runtime integration with Primo Search API, which may introduce latency or complexity. (We can mitigate this by implementing a caching strategy similar to that in Bento.)
```
JPrevost, line 130: `which may` should probably be `which will`. We already know Primo API is extremely slow.

docs/architecture-decisions/0003-surface-primo-cdi-records-in-results.md
```diff
@@ -0,0 +1,146 @@
...
+in TIMDEX API, we could facilitate computational access to a massive corpus of data.
+
+Unfortunately, this approach is not feasible for many reasons:
+
+- **Cost**: CDI contains over 5 billion records. Harvesting and storing these records would be impractical and expensive, both in terms of financial and compute resources.
+- **Performance**: Expanding TIMDEX API at such a scale is likely to dramatically reduce the efficiency of our OpenSearch index.
+- **Data availability**: Because Primo does not expose CDI records in OAI-PMH, we would need to harvest using the Primo Search API, making the process needlessly complex and perhaps impossible.
+- **Licensing**: Harvesting CDI records for TIMDEX likely has licensing implications. Ex Libris seems to discourage the practice, as Primo does not provide OAI-PMH support, and the Search API caps records per request at 5,000 via the [`offset` parameter](https://developers.exlibrisgroup.com/primo/apis/docs/primoSearch/R0VUIC9wcmltby92MS9zZWFyY2g=/#output:~:text=Note%3A%20The%20Primo%20search%20API%20has%20a%20hardcoded%20offset%20limitation%20parameter%20of%205000.).
+
+### Display separate result streams in tabbed views
```
JPrevost, line 35: I believe there is a desire from the UX side to have tabs for "primo" and "timdex" even if we figure out how to merge all the results. We may want to update our decision to reflect having multiple tabs ("everything (which is what I believe the Orchestrator approach would provide), "primo (probably Alma + CDI)", "timdex (everything not in the primo tab)"

docs/architecture-decisions/0003-surface-primo-cdi-records-in-results.md
```diff
@@ -0,0 +1,146 @@
...
+The Libraries' unified search strategy calls for a discovery interface that surfaces results from
+both Primo Central Discovery Index (CDI) and Alma (via TIMDEX), replacing the current [Bento UI](https://github.com/MITLibraries/bento).
+In Bento, Alma and CDI results are displayed in separate boxes. The unified interface would
+interleave CDI and TIMDEX records in the same results list.
+
+## Options considered
+
+### Harvest Primo CDI data
+
+We considered adding a new Primo harvester to our ETL architecture to ingest CDI data into TIMDEX
```
JPrevost, line 20: I'd come down stronger with "This is not feasible" (or similar) as the first sentence.

docs/architecture-decisions/0003-surface-primo-cdi-records-in-results.md
```diff
@@ -0,0 +1,146 @@
...
+
+### Display separate result streams in tabbed views
+
+This option would essentially be a different take on the Bento design. On the results page, a user
+could tab between Alma results (labeled 'Books', 'MIT Catalog', etc.) and CDI results ('Articles').
+
+While arguably an improvement on Bento, this design does not deliver the combined Alma/CDI results
+view as envisioned in the unified UI. 
+
+### Implement external search orchestrator
```
JPrevost, line 43: Not for this document, but we'll likely need a separate ADR for the orchestrator if we agree that is the best path forward as I suspect there are a few solid directions to go (Rails, lambda, lambda feeding OpenSearch, etc)

JPrevost (commented):
I'd love to meet as a team to discuss these options now that we have this document as a starting point.

docs/architecture-decisions/0003-surface-primo-cdi-records-in-results.md
jazairi: Agreed. I talked a bit in Slack about deferred architectural decisions for the orchestrator, but I'm not sure it translated into the ADR.

JPrevost: I hadn't caught up on Slack (still haven't fully!). We don't need to update this ADR to note there will be a different ADR with more details on whatever we choose... I was just saying it out loud :)

jazairi: Got you. It just felt useful to me to note it in the ADR, because I think at times we start developing before we're done documenting decisions.

JPrevost: Ah... that's a very good point.

jazairi:
We've decided to postpone this decision pending further exploration of the problem space.

PR #290, opened 2025-11-26
Improve 'all' tab pagination to handle edge cases

jazairi:

The zipper merge we implemented naively queries n/2 results from each API and interleaves them, where n is the per-page value. This works if both APIs return many results, but it can cause problems in smaller, unbalanced result sets.

For example, the query term `doc edgerton` returns 50 Primo results and 4 TIMDEX results. Page 1 only shows 14 results (4 TIMDEX and 10 Primo), and each subsequent page returns only 10 (all Primo).

#### Relevant ticket(s):

- [USE-179](https://mitlibraries.atlassian.net/browse/USE-179)

This implements more sophisticated logic that first checks the number of hits returned by each API and passes that, along with the pagination information, to a Merged Search Paginator class. This service object develops a 'merge plan', calculates API offsets, and merges the results for each page.

Queries on the 'all' tab now fetch twice from each API: once to determine the total number of hits for the Merged Search Paginator then again to fetch results at the appropriate offset. While hardly ideal, this was the only option I could figure to avoid losing results. I limited these extra calls to queries beyond page 1, which is the only case where they are needed.

* We now clear cache before each search controller test. This was done to avoid odd test behavior, but I ran the suite 50 times without any issues, so it might be excessively cautious.
* The search controller continues to grow with this new logic. I tried to split things into multiple helper methods, so if we want to move more things to service objects later, it might be easier to do so.
* A failing cassette has been replaced with a mock.

##### Accessibility

- [x] This PR contains no changes to the view layer.

##### New ENV

- [x] ENV has not changed.

This is a pretty unwieldy changeset, so please reach out if you have questions!

##### Code

added technical debt.

##### Documentation

##### Testing

[USE-179]: https://mitlibraries.atlassian.net/browse/USE-179?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/models/merged_search_service.rb
```diff
@@ -0,0 +1,241 @@
...
+  # @param enhanced_query [Hash] query hash produced by `Enhancer`
+  # @param active_tab [String] the currently active tab (e.g. 'all')
+  # @param cache [Object] optional cache store responding to `read`/`write` (defaults to `Rails.cache`)
+  # @param primo_fetcher [#call] optional callable used to fetch Primo results; should accept `offset:, per_page:, query:`
+  # @param timdex_fetcher [#call] optional callable used to fetch TIMDEX results; should accept `offset:, per_page:, query:`
+  def initialize(enhanced_query:, active_tab:, cache: Rails.cache, primo_fetcher: nil, timdex_fetcher: nil)
+    @enhanced_query = enhanced_query
+    @active_tab = active_tab
+    @cache = cache
+    @primo_fetcher = primo_fetcher || method(:default_primo_fetch)
```
JPrevost: It's unclear when/why we are injecting versus using the default methods.

The controller seems to be injecting so I'm not sure when the defaults are used.

app/models/merged_search_paginator.rb
```diff
@@ -0,0 +1,73 @@
...
+      else
+        break
+      end
+      i += 1
+    end
+    [primo_offset, timdex_offset]
+  end
+
+  # Merges two result arrays according to the merge plan
+  def merge_results(primo_results, timdex_results)
```
JPrevost: This feels more like part of the merge service than the paginator service.

app/models/merged_search_service.rb
```diff
@@ -0,0 +1,241 @@
...
+  # @param per_page [Integer]
+  # @param deeper [Boolean] whether this was a deeper-page flow
+  # @return [Hash] response with :results, :errors, :pagination, :show_primo_continuation
+  def assemble_all_tab_result(paginator, primo_data, timdex_data, current_page, per_page, deeper: false)
+    primo_total = primo_data[:hits] || 0
+    timdex_total = timdex_data[:hits] || 0
+
+    merged = paginator.merge_results(primo_data[:results] || [], timdex_data[:results] || [])
+    errors = combine_errors(primo_data[:errors], timdex_data[:errors])
+    pagination = Analyzer.new(@enhanced_query, timdex_total, :all, primo_total).pagination
```
JPrevost: I'm not sure if this is a bug that can happen in this application, or just in the console.

But here's a sequence of console commands that lead to an unexpected result for this pagination value (tldr; it uses the per_page for the app whereas all of these methods allow you to pass any value you want... maybe they shouldn't accept values and just use the per_page of the app to avoid having someone override something at some point and have pagination blow up?)

```
params = {q: 'hello'}
q = Enhanger.new(params)
s = MergedSearchService.new(enhanced_query: q, active_tab: 'all')
s.fetch(page: 1, per_page: 1)
```

Pagination will then be:
```
pagination: {hits: 162397, per_page: 20, start: 1, end: 20, next: 2}
```

Which has the correct offsets and next, but the wrong per_page.

test/models/merged_search_service_test.rb
```diff
@@ -0,0 +1,202 @@
+require 'test_helper'
+require 'ostruct'
```
JPrevost: Are we only using OpenStruct in testing or has it ended up on the running app as well?

Zero concerns if it's just in tests, minor nervousness if it is in the main codebase because it has the _potential_ to cause unexpected slow downs. Mostly we'd just want to keep an eye on it, not avoid its use.

JPrevost (commented):
I still need to work through the test suites for the two new models but wanted to share my feedback so far as I need to step away from the review for a bit and thought in-progress feedback may be helpful rather than waiting until I can get back to it.

app/models/merged_search_service.rb
```diff
@@ -0,0 +1,235 @@
...
+  end
+
+  # Persist per-query totals to cache(s).
+  #
+  # The method writes to the injected cache (if available) and to
+  # `Rails.cache`. Additional marker keys are written to improve test
+  # discoverability for stores that are probed with `read_matched`.
+  #
+  # @param totals [Hash] { primo: Integer, timdex: Integer }
+  def write_cached_totals(totals)
```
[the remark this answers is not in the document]
jazairi: Honestly, this method can probably be simplified a lot. For example, one cache write was for debugging while I was writing the feature, and I thought it might be useful for future debugging. Looking at this with fresh eyes, though, I don't think it's necessary..

I've got a version that it is hopefully clearer on intent, which I'll push up once I've gone through the remaining feedback.

app/models/merged_search_service.rb
```diff
@@ -0,0 +1,235 @@
...
+      primo_data, timdex_data = parallel_fetch(offset: 0, per_page: per_page)
+
+      totals = { primo: primo_data[:hits].to_i, timdex: timdex_data[:hits].to_i }
+      write_cached_totals(totals)
+
+      paginator = build_paginator_from_totals(totals, current_page, per_page)
+
+      results = assemble_all_tab_result(paginator, primo_data, timdex_data, current_page, per_page)
+
+      return results
```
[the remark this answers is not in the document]
jazairi: If you think it would improve readability, that's good enough for me!

app/models/merged_search_paginator.rb
jazairi: Probably. It started in the paginator service because it predated the merge service, and I left it there out of laziness. I agree that it makes more sense in the merge service, though. I'll explore what that might entail (I think it's just rewriting some tests).

test/models/merged_search_service_test.rb
jazairi: We're just using it in testing.

app/models/merged_search_service.rb
```diff
@@ -0,0 +1,319 @@
...
+  # Time to live value for cache expiration.
+  TTL = 12.hours
+
+  # Initialize a new MergedSearchService.
+  #
+  # The service supports dependency injection for the two fetchers to keep
+  # calling code simple and to make testing deterministic.
+  #
+  # Usage patterns:
+  # - Default: callers omit fetcher arguments and the service uses `Rails.cache`
```
JPrevost: It's still a bit unclear to me if we ever use the default path or if we are always injecting...

app/models/merged_search_service.rb
JPrevost: It's still unclear to me if we ever use the default path or if we are always injecting... it seems like the Controller and Tests both inject so is default never used?

test/test_helper.rb
```diff
@@ -42,3 +42,18 @@ class TestCase
...
 end
+
+# Use a memory-backed `Rails.cache` for tests and clear it before each test to avoid cross-test
+# leakage. This simplifies tests and emoves the need to inject a cache store into most specs, but
+# a shared cache can still cause uses with parallel fetching, as we use in the Merged Search
+# Service.
+#
+# If you hit flaky tests related to cache  access under concurrency, consider injecting a per-test
+# cache or avoiding shared mutable state in the threaded path.
+Rails.cache = ActiveSupport::Cache::MemoryStore.new
```
JPrevost: It might be better to change this value in the config rather than overriding it here?

https://github.com/MITLibraries/timdex-ui/blob/main/config/environments/test.rb#L26

JPrevost (approved):
:shipit:

PR #377, opened 2026-04-03
USE-404: Implements Matomo Analytics in USE UI

djanelle-mit:
#### Developer
We wanted to implement our Analytics strategy in USE UI. While we originally tried using Tag Manager's UI, we ran into problems with our single page application that prevented that path. This path implements helper functions inside USE UI that wraps the default `_paq.push` method of sending Events to Matomo, making it straightforward to track Events in code.

Using `_paq.push` to send Events to Matomo co-locates these in the Matomo dashboard alongside Tag Manager's events. If we decide to use Tag Manager in the future, the tracking done here can live beside any Tag Manager events without issue.

This work also includes helper functions that provide dynamic data to our tracking events (which tab is active, what page of results we're on, etc.).

Finally, this work implements the items from our first milestone of our Analytics Strategy for USE. Events were created for all intended events, and content tracking was introduced as an experiment to get more robust data about what type of content users interact with.

##### Accessibility

##### New ENV

- [x] ENV has not changed.

##### Code

added technical debt.

##### Documentation

##### Testing

### DIFF
--- a/app/javascript/matomo_tracking.js
+++ b/app/javascript/matomo_tracking.js
@@ -12,7 +12,7 @@
 // Tracking also works if placed on a container. The script will look
 // for any interactive elements inside the container.
 //
-// Interactive elemenets are defined as elements: a, button, input, select, textarea
+// Interactive elements are defined as elements: a, button, input, select, textarea
 //
 // Examples:
 //   <a href="/file.pdf" data-matomo-click="Downloads, PDF Click, My Paper">Download</a>
@@ -39,7 +39,7 @@
 // must be registered on `window.MatomoHelpers` (see bottom of this file).
 // Multiple tokens in one segment are supported.
 //
-// Convention is to only use these in the "Name" segement to provide more context.
+// Convention is to only use these in the "Name" segment to provide more context.
 // Avoid using inside Category or Action to improve the hierarchy of Matomo dashboards.
 //
 // Examples:
@@ -168,6 +168,17 @@ document.querySelectorAll("[data-matomo-seen]").forEach((el) => {
 // ---------------------------------------------------------------------------
 // Matomo native content tracking
 // ---------------------------------------------------------------------------
+
+// Core Matomo includes "Content Tracking" 
+// You can use these attributes to track content impressions and interactions.
+// See documentation for details (https://matomo.org/faq/how-to/how-do-i-markup-content-for-content-tracking/)
+// This file only ensures that these attributes are processed for asynchronously-inserted content.
+// 
+// Attributes for content tracking (native to Matomo):
+// * data-track-content - Defines a content block to track
+// * data-content-name - Gives a name for the content block (appears in Matomo Dashboard)
+// * data-content-piece - Names a piece of content to track interactions on (appears in Matomo Dashboard)
+// * data-content-target - Specifies the target of the content interaction (appears in Matomo Dashboard)
 
 // Matomo's built-in content tracking (data-track-content / data-content-name /
 // data-content-piece) only scans the DOM at page load. For content injected
--- a/app/javascript/matomo_tracking.js
+++ b/app/javascript/matomo_tracking.js
@@ -44,7 +44,7 @@
 //
 // Examples:
 //   <h2 data-matomo-seen="Search, Results Found, Tab: {{getActiveTabName}}">...</h2>
-//   <a data-matomo-click="Nav, {{getActiveTabName}} Link Click">...</a>
+//   <a data-matomo-click="Nav, Link Click, Link: {{getElementText}}">...</a>
 
 // ---------------------------------------------------------------------------
 // Shared helper
--- a/app/javascript/matomo_tracking.js
+++ b/app/javascript/matomo_tracking.js
@@ -1,0 +1,275 @@
+// BACKGROUND
+// This file implements helper functions that make it very straigtforward to track Matomo events in code.
+// These functions use matomo's _paq.push() API to send events directly to Matomo.
+// This does NOT use Tag Manager, but works in core matomo or Tag Manager environments.
+//
+// CLICK TRACKING
+// Add `data-matomo-click="Category, Action, Name"` to any element to track
+// clicks as Matomo events. The Name segment is optional. We have a convention 
+// of using semicolons inside Name to divide multiple values.
+//
+// This tracking can be placed directly on an interactive element.
+// Tracking also works if placed on a container. The script will look
+// for any interactive elements inside the container.
+//
+// Interactive elemenets are defined as elements: a, button, input, select, textarea
+//
+// Examples:
+//   <a href="/file.pdf" data-matomo-click="Downloads, PDF Click, My Paper">Download</a>
+//   <button data-matomo-click="Search, Boolean Toggle">AND/OR</button>
+//
+// Event delegation on `document` means this works for elements loaded
+// asynchronously (Turbo frames, content-loader, etc.) without re-binding.
+//
+// SEEN TRACKING
+// Add `data-matomo-seen="Category, Action, Name"` to any element to fire a
+// Matomo event when that element becomes visible in the viewport. The Name
+// segment is optional. We have a convention of using semicolons inside Name 
+// to divide multiple values. Each element fires at most once per page load.
+// Works for elements present on initial page load and for elements injected
+// later by Turbo frames or async content loaders.
+//
+// Examples:
+//   <div data-matomo-seen="Impressions, Result Card, Alma">...</div>
+//   <a data-matomo-seen="Promotions, Banner Shown">...</a>
+//
+// DYNAMIC VALUES
+// Wrap a helper name in double curly braces anywhere inside a segment to have
+// it replaced with the return value of that function at tracking time. Helpers
+// must be registered on `window.MatomoHelpers` (see bottom of this file).
+// Multiple tokens in one segment are supported.
+//
+// Convention is to only use these in the "Name" segement to provide more context.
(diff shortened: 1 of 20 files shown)


app/views/layouts/_global_alert.html.erb
```diff
@@ -0,0 +1,9 @@
+<% if ENV['GLOBAL_ALERT'] %>
+  <div class="wrap-notices info layout-band">
+    <div class="wrap-notice">
+      <div class="alert alert-global" data-matomo-click="Banner, Link Engaged, Link: {{getElementText}}">
```
matt-bernhardt, line 4: Non-blocking comment: I think I see why this file is here - we're copying this partial from the theme gem into this repo so that we can add the `data-matomo-click` attribute here, right? When i diff this file against the version in the gem, that's the only change.

This makes sense to me, so I'm not objecting to the change. The way it's introduced in the PR (all in one commit) is a little harder to follow than it could be, though. My preference would be for one commit to drop the existing file in place (with a commit message clarifying that no changes are being made), with an addition like this introduced in a second commit.

This is a little fastidious, I recognize - it's also easy enough to just diff the two files against each other as I've done here.

app/javascript/matomo_tracking.js
```diff
@@ -0,0 +1,275 @@
+// BACKGROUND
+// This file implements helper functions that make it very straigtforward to track Matomo events in code.
+// These functions use matomo's _paq.push() API to send events directly to Matomo.
+// This does NOT use Tag Manager, but works in core matomo or Tag Manager environments.
+//
+// CLICK TRACKING
+// Add `data-matomo-click="Category, Action, Name"` to any element to track
+// clicks as Matomo events. The Name segment is optional. We have a convention 
+// of using semicolons inside Name to divide multiple values.
```
matt-bernhardt, line 9: Question: So with the semicolon convention we're adopting here, when I see a value like:
`Navigation, Pagination Engaged, Link: {{getElementText}}; Current Page: {{getCurrentResultsPage}}`
The semicolons aren't used in a functional way - they just influence how a single element is reported in the Matomo interface? When I look in Matomo, I see data like this:

Image

I wanted to make sure that I'm not missing functionality like a semicolon results in two event actions being fired - this is purely a readability concern within the Matomo reporting UI?

app/javascript/matomo_tracking.js
```diff
@@ -0,0 +1,275 @@
...
+// it replaced with the return value of that function at tracking time. Helpers
+// must be registered on `window.MatomoHelpers` (see bottom of this file).
+// Multiple tokens in one segment are supported.
+//
+// Convention is to only use these in the "Name" segement to provide more context.
+// Avoid using inside Category or Action to improve the hierarchy of Matomo dashboards.
+//
+// Examples:
+//   <h2 data-matomo-seen="Search, Results Found, Tab: {{getActiveTabName}}">...</h2>
+//   <a data-matomo-click="Nav, {{getActiveTabName}} Link Click">...</a>
```
matt-bernhardt: Question: Does this example violate the convention stated above about using dynamic values in Actions?

app/javascript/matomo_tracking.js
```diff
@@ -0,0 +1,275 @@
...
+
+// ---------------------------------------------------------------------------
+// Matomo native content tracking
+// ---------------------------------------------------------------------------
+
+// Matomo's built-in content tracking (data-track-content / data-content-name /
+// data-content-piece) only scans the DOM at page load. For content injected
+// asynchronously (e.g. by the content-loader Stimulus controller), we must
+// manually notify Matomo by calling trackContentImpressionsWithinNode on the
+// newly-added node.
```
matt-bernhardt, line 187: **Requested change:**
I see this documentation referencing data-track-content but in a handful of templates we're using data-content-track instead? I do see the track-content variant in at least three templates, but it feels like we're using the wrong value on the _results and _results_primo partials?

app/views/search/_result.html.erb
```diff
@@ -1,4 +1,4 @@
-<li class="result use" data-record-id="<%=result[:identifier]%>">
+<li class="result use" data-record-id="<%= result[:identifier] %>" data-matomo-seen="Results, Result Seen, Tab: {{getActiveTabName}}" data-content-track data-content-name="Result">
```
matt-bernhardt: **Requested change:**
Should this set of data attributes include `data-track-content` instead of `data-content-track`?

app/views/search/_result_primo.html.erb
```diff
@@ -1,11 +1,11 @@
-<li class="result primo" data-record-id="<%=result[:identifier]%>">
+<li class="result primo" data-record-id="<%= result[:identifier] %>" data-matomo-seen="Results, Result Seen, Tab: {{getActiveTabName}}" data-content-track data-content-name="Result">
```
matt-bernhardt: **Requested change:**
Should this set of data attributes include `data-track-content` instead of `data-content-track`?

app/javascript/matomo_tracking.js
```diff
@@ -0,0 +1,275 @@
...
+  // (or the element itself) that has a data-matomo-click attribute.
+  const el = event.target.closest("[data-matomo-click]");
+  // If no such element exists in the ancestor chain, ignore this click.
+  if (!el) return;
+
+  // Only fire when the click originated from an interactive element (link,
+  // button, or form control). This allows data-matomo-click to be placed on
+  // a container and track only meaningful interactions within it, ignoring
+  // clicks on surrounding text, padding, or decorative children.
+  const interactive = event.target.closest("a, button, input, select, textarea");
```
matt-bernhardt, line 108: Comment:
I agree with having a filter for interactable elements in this position, and I think the set of elements we're going with is fine for the moment.

What I wonder is how this intersects with our a11y best practices, and whether a list like this needs to be accounted for in our interaction and a11y approach.

matt-bernhardt (changes requested):
I think there are two specific changes needed, assuming I've understood what matomo_tracking.js is doing. It seems to be looking for a `data-track-content` attribute, which many of the partials are adding, but two (_result and _result_primo) are adding `data-content-track` instead? I think those two partials need to have their attribute names updates?

---
A more general maybe-request I have is that matomo_tracking might benefit from a clearer specification of "these are the data attributes supported by this block of code:" Some are listed here and there, and I think the full list is discoverable if I review the diff, but it might help future-us to have a clear list of "this is what's supported":

- used by core Matomo
  - data-content-name
  - data-content-piece
  - data-track-content

- Added by this library:
  - data-matomo-click
  - data-matomo-seen

---
The other comments in here are either process notes, or me making sure that I'm not misinterpreting the functionality I think is present (how semicolons are treated).

I'm also turning over in my head how we might write tests for this functionality - because I think a test might reveal the `data-content-track` / `data-track-content` thing (which feels like an easy place for a regression to creep in later). I don't want to block merging this on that kind of lack of tests, though. The only needed change is to those two partials.

app/javascript/matomo_tracking.js
djanelle-mit: Yep, this was entirely to have a consistent delimiter between the "Name" key value pairs. I could see us needing to parse these with something like Excel at some point in the future (given how flimsy Matomo's dashboards are) so I wanted to make sure we're consistent about applying the delimiter.

app/javascript/matomo_tracking.js
djanelle-mit: Yes, I'll update this.

app/views/search/_result_primo.html.erb
djanelle-mit: 🤦 Yes. Yes it should. It bothers me to no end that all of the other content tracking attributes are data-content-XYZ but this one isn't. I'll update this plus the other instance.

matt-bernhardt: In fairness, I can't get _too_ upset about that inconsistency without also remembering that we've used `main-content` and `content-main` in conjunction because we've also not been consistent...

But yes - that bugs me too.

app/views/layouts/_global_alert.html.erb
djanelle-mit: Yes, you're spot on here, I added this purely to add the tracking attributes. This PR was a bit of a journey given all the experiments, but I'll keep this in mind (adding a commit JUST to add the file from the theme) in future PRs!

app/javascript/matomo_tracking.js
djanelle-mit: Would you be able to expand a bit more here? Are you worried about this list not being encompassing enough and tracking on other elements outside of this list being missed because of it?

app/javascript/matomo_tracking.js
djanelle-mit: Updated all references to use `data-track-content` instead of `data-content-track`

app/javascript/matomo_tracking.js
matt-bernhardt: Yeah, I'm wondering if we'll end up wanting to track interactions with other elements in a UI, and end up wondering why records aren't being recorded because the UI element ends up using something like `div role="button"`.

I kind of suspect a few things might also be true, though:
- That we'd end up isolating this as the bottleneck and be able to expand the list when needed
- I suspect `<div role="button">` might also be an a11y ant-pattern, so it might be okay to introduce friction against bad ideas

Ultimately, I don't think any changes are needed right now - this is just something I'm wondering about as we go forward.

matt-bernhardt (approved):
This looks good - thanks for adding the attribute list, and the reference to the Matomo documentation. That should definitely help future-me.

Thanks also for swapping the data attributes on those two partials. With that, I think we're ready to go here.

:shipit:

PR #383, opened 2026-04-22
Adds additional tabs for timdex sources

JPrevost:

* All sources should exist in at least one tab

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/USE-517

* Adds additional tabs for timdex sources
* Adds additional blurbs for the new tabs

##### Accessibility

##### New ENV

- [ ] ENV has not changed.

##### Code

added technical debt.

##### Documentation

##### Testing

### DIFF
--- a/app/helpers/results_helper.rb
+++ b/app/helpers/results_helper.rb
@@ -1,27 +1,39 @@
 module ResultsHelper
+  # Descriptions for each tab. HTML entries use raw anchor tags (all URLs are
+  # hardcoded external links, so no XSS risk) and must be marked html_safe.
+  TAB_DESCRIPTIONS = {
+    'all' => 'Search across all MIT Libraries systems',
+    'cdi' => 'Articles, reviews, book chapters, and more from ' \
+             '<a href="https://mit.primo.exlibrisgroup.com/discovery/search?vid=01MIT_INST:MIT&lang=en">' \
+             'Articles, Books & More</a>'.html_safe,
+    'alma' => 'Books, e-books, journals, streaming and physical media, and more from ' \
+              '<a href="https://mit.primo.exlibrisgroup.com/discovery/search?vid=01MIT_INST:MIT&lang=en">' \
+              'Articles, Books & More</a>'.html_safe,
+    'timdex_alma' => 'Books, e-books, journals, streaming and physical media, and more',
+    'primo' => 'Articles, books, chapters, streaming and physical media, and more',
+    'aspace' => 'Finding aids for archival and unique primary source materials at MIT from ' \
+                '<a href="https://archivesspace.mit.edu">Archives & Manuscripts</a>'.html_safe,
+    'timdex' => 'Digital collections, images, documents, and more from MIT Libraries',
+    'website' => 'Events, news, services, and research guides from the ' \
+                 '<a href="https://libraries.mit.edu">Library Website</a> and ' \
+                 '<a href="https://libraries.mit.edu/experts/">Research Guides</a>'.html_safe,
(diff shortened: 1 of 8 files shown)


test/helpers/results_helper_test.rb
```diff
@@ -21,37 +21,10 @@ class ResultsHelperTest < ActionView::TestCase
...
-    assert_equal description, tab_description
-
-    params[:tab] = 'timdex'
-    description = 'Digital collections, images, documents, and more from MIT Libraries'
-    assert_equal description, tab_description
-
-    params[:tab] = 'website'
-    description = 'Information about the library: events, news, services, and more'
-    assert_equal description, tab_description
+    ResultsHelper::TAB_DESCRIPTIONS.each do |tab, expected|
```
matt-bernhardt, line 24: I like the efficiency of this test, but I wonder whether we should have a test for the behavior when a value outside that `TAB_DESCRIPTIONS` construct is presented?

More theoretically, I get that this test makes things more robust because we don't need to manually keep the list of test values in sync with the content of the helper method. This feels like a win - but I'm not sure where the line is where we just end up testing that `.each()` does what we expect.

app/helpers/results_helper.rb
```diff
@@ -1,27 +1,34 @@
 module ResultsHelper
+  # Descriptions for each tab. HTML entries use raw anchor tags (all URLs are
+  # hardcoded external links, so no XSS risk) and must be marked html_safe.
+  TAB_DESCRIPTIONS = {
```
matt-bernhardt, line 4: I like this change in pattern a lot. 🍻

app/helpers/results_helper.rb
```diff
@@ -1,27 +1,34 @@
...
-    when 'aspace'
-      'Archives, manuscripts, and other unique materials related to MIT'
-    when 'timdex'
-      'Digital collections, images, documents, and more from MIT Libraries'
-    when 'website'
-      'Information about the library: events, news, services, and more'
-    else
+    TAB_DESCRIPTIONS.fetch(params[:tab]) do
       Rails.logger.error "Unknown tab parameter in `tab_description` helper: #{params[:tab]}"
+      ''
```
matt-bernhardt, line 36: Making sure I understand - this is the default value in case `tab_description` gets presented a value that we don't expect, right? So if there's a regression or mismatch somehow, we're not presenting gibberish, just an empty string.

matt-bernhardt (approved):
I have a non-blocking request for a different test type, but it isn't something I'm going to frame as a full request for changes - if there's something in this that I'm missing, I'm happy for this to merge as-is (the type of test I'm talking about didn't exist before, so this isn't a regression)

djanelle-mit:
From my perspective, this looks good to go. Having a const to control the max viewable tabs makes sense. Since the truncation of tabs happens on pretty wide screens now, I should probably adjust the Tablet+ breakpoints to have a more traditional dropdown menu for the "more" overflow instead of the full width menu.

We can create a follow-up ticket for that though, and definitely shouldn't block things.

app/helpers/results_helper.rb
JPrevost: Correct. Previous to this we were inadvertently returning `true` as the logging statement is apparently truthy.

test/helpers/results_helper_test.rb
JPrevost: I definitely agree we should confirm something outside of the expected list returns blank. I'll confirm that test exists or add one.

JPrevost: I've added additional tests as suggested here. Thanks for pointing out we were missing edge cases!

matt-bernhardt (approved):
Yep, I think this looks good. Thanks!

PR #386, opened 2026-05-07
USE-457: Swapping material type and source locations

djanelle-mit:

* stakeholders wanted eyebrows to be based on format/content_type instead
  of source
* stakeholders wanted source to be mapped to user friendly strings with
  links to source systems

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/USE-587
* https://mitlibraries.atlassian.net/browse/USE-457

* Updates eyebrow mappings to be based on format/content_type instead of source
* Adds source mappings to user friendly strings with links to source systems

* Visited link to source may need a UI pass to make it look better

##### Accessibility

##### New ENV

- [ ] ENV has not changed.

##### Code

added technical debt.

##### Documentation

##### Testing

### DIFF
--- a/app/models/normalize_timdex_record.rb
+++ b/app/models/normalize_timdex_record.rb
@@ -1,5 +1,19 @@
 # Transforms a TIMDEX result into a normalized record.
 class NormalizeTimdexRecord
+  # Maps raw TIMDEX source names to [display_name, url] pairs.
+  # Add new sources here — both fields stay in sync automatically.
+  SOURCES = {
+    'DSpace@MIT' => ['MIT Open Scholarship (DSpace@MIT)', 'https://dspace.mit.edu/'],
+    'LibGuides' => ['Research Guides', 'https://libguides.mit.edu/'],
+    'MIT ArchivesSpace' => ['Archives & Manuscripts', 'https://archivesspace.mit.edu'],
+    'OpenGeoMetadata GIS Resources' => ['Open Geospatial Consortium', 'https://opengeometadata.org/'],
+    'MIT GIS Resources' => ['MIT Geospatial Data', 'https://geodata.libraries.mit.edu/'],
+    'Research Databases' => ['Research Databases', 'https://libguides.mit.edu/az/databases'],
+    'MIT Libraries Website' => ['Library Website', 'https://libraries.mit.edu/'],
+    'MIT Alma' => ['MIT Libraries Catalog',
+                   "#{ENV.fetch('MIT_PRIMO_URL')}/discovery/search?vid=#{ENV.fetch('PRIMO_VID')}&lang=en"]
+  }.freeze
+
   def initialize(record, query)
     @record = record
     @query = query
@@ -13,6 +27,7 @@ class NormalizeTimdexRecord
       creators:,
       eyebrow:,
       source:,
+      source_url:,
       year:,
       format:,
       links:,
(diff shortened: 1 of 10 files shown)


app/models/normalize_timdex_record.rb
```diff
@@ -1,5 +1,19 @@
...
+    'DSpace@MIT' => ['MIT Open Scholarship (DSpace@MIT)', 'https://dspace.mit.edu/'],
+    'LibGuides' => ['Research Guides', 'https://libguides.mit.edu/'],
+    'MIT ArchivesSpace' => ['Archives & Manuscripts', 'https://archivesspace.mit.edu'],
+    'OpenGeoMetadata GIS Resources' => ['Open Geospatial Consortium', 'https://opengeometadata.org/'],
+    'MIT GIS Resources' => ['MIT Geospatial Data', 'https://geodata.libraries.mit.edu/'],
+    'Research Databases' => ['Research Databases', 'https://libguides.mit.edu/az/databases'],
+    'MIT Libraries Website' => ['Library Website', 'https://libraries.mit.edu/'],
+    'MIT Alma' => ['MIT Libraries Catalog',
+                   "#{ENV.fetch('MIT_PRIMO_URL')}/discovery/search?vid=#{ENV.fetch('PRIMO_VID')}&lang=en"]
+  }.freeze
```
matt-bernhardt, line 15: Non-blocking comment...

I wonder about the comparison between this approach (key-value pairs in the top of a normalizer model) versus the format vocabular model (separate class, callable by either normalizer, also holding key-value pairs)

This is probably fine - I see that we're declaring all Primo sources to be just the new name for Primo, and not separating CDI from Alma records. This is the only normalizer that would need a structure like this. (I also  see we're declaring an Alma value here, but I don't think we're searching these records in USE? Maybe in one of the other instances?)

Ultimately, I'd like similar concerns to be handled similarly - whether that's this pattern or the vocabulary pattern doesn't feel like a deeply held preference. But that doesn't have to happen in this PR.

app/views/search/_result.html.erb
```diff
@@ -1,6 +1,6 @@
 <li class="result use" data-record-id="<%= result[:identifier] %>" data-matomo-seen="Results, Result Seen, Tab: {{getActiveTabName}}" data-track-content data-content-name="Result">
   <div class="result-content">
-    <% if multi_source_tab? && result[:eyebrow].present? %>
```
matt-bernhardt, line 3: Does this change (also the removal from the _result_primo template) orphan the multi_source_tab? helper? I don't see another reference to it in `app/` from a brief search

matt-bernhardt (approved):
I'm happy to approve this, but I suspect that there's a helper being orphaned that should also be removed (either in this PR or via a future piece of work that's ticketed).

My other comment is non-blocking navel-gazing, but I'd like to discuss in Slack after this merges.

:shipit:

app/models/normalize_timdex_record.rb
JPrevost: The Alma value here is for when we use TIMDEX for searching Alma, so it remains solely focused on TIMDEX normalization. I believe Vocabularies are used cross-API but I could be mistaken.

It's an intriguing idea to extract this into a very simple ruby class, but I think I lean towards not doing that at this time.

app/views/search/_result.html.erb
JPrevost: Ah, great catch. I'll confirm it is no longer used and remove it if so.

JPrevost: Confirmed this was no use anywhere and removed the helper method and tests.

PR #395, opened 2026-05-21
Implement natural language search opt-in

jazairi:

We want users to be able to decide whether to use
hybrid (termed 'natural language search' in the
public UI) or lexical search. USE-577 implemented
the frontend work on this feature; this ticket
implements the controller layer.

#### Relevant ticket(s):

- [USE-576](https://mitlibraries.atlassian.net/browse/USE-576)
- [USE-577](https://mitlibraries.atlassian.net/browse/USE-577)

This adds natural language search opt-in cookie,
which is toggled via a stimulus controller. When
the cookie is set, hybrid search is the default
query mode. When it is not set, the query mode
falls back to lexical search (or other default
mode specified by env).

This behavior can be overridden by supplying a
`queryMode` param manually. For example, if a user has the NLS cookie set but adds `queryMode=semantic` to their query string, the search will execute
in semantic mode, not hybrid.

If NLS is not available for a given tab, a
message displays to notify the user of the
limitation.

Tests for this feature are stored in a separate
file. This is different from how we usually do
things, but there are so many conditions to test
that I did not want to pollute the search
controller test file further.

##### Accessibility

##### New ENV

- [x] ENV has not changed.

##### Code

added technical debt.

##### Documentation

##### Testing

[USE-576]: https://mitlibraries.atlassian.net/browse/USE-576?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[USE-577]: https://mitlibraries.atlassian.net/browse/USE-577?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,6 +1,8 @@
 class ApplicationController < ActionController::Base
   helper Mitlibraries::Theme::Engine.helpers
 
+  before_action :set_natural_language_search_optin
+
   # Set active tab based on params (no persistent cookie). This intentionally
   # avoids storing the user's last-used tab in a cookie per UXWS request.
   def set_active_tab
@@ -30,5 +32,9 @@ class ApplicationController < ActionController::Base
 
   def valid_tab?(tab)
     all_tabs.include?(tab)
+  end
+
+  def set_natural_language_search_optin
+    @natural_language_search_optin = cookies[:natural_language_search_optin] == 'true'
   end
 end
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -16,7 +16,7 @@ class SearchController < ApplicationController
     # inject session preference for boolean type if it is present
     params[:booleanType] = cookies[:boolean_type] || 'AND'
 
-    # inject queryMode from cookie if not provided in URL params
+    # inject hybrid queryMode if opted-in and no queryMode param provided
     params[:queryMode] = 'hybrid' if params[:queryMode].blank? && cookies[:natural_language_search_optin] == 'true'
 
     @enhanced_query = Enhancer.new(params).enhanced_query
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -16,7 +16,11 @@ class SearchController < ApplicationController
     # inject session preference for boolean type if it is present
     params[:booleanType] = cookies[:boolean_type] || 'AND'
 
+    # inject queryMode from cookie if not provided in URL params
+    params[:queryMode] = 'hybrid' if params[:queryMode].blank? && cookies[:natural_language_search_optin] == 'true'
+
     @enhanced_query = Enhancer.new(params).enhanced_query
+    @show_nls_warning = show_nls_warning?
 
     # Load GeoData results if applicable
     if Feature.enabled?(:geodata)
@@ -441,5 +445,11 @@ class SearchController < ApplicationController
     return true if ActiveSupport::SecurityUtils.secure_compare(params[:format_token], ENV.fetch('FORMAT_TOKEN', ''))
 
     false
+  end
+
+  # show_nls_warning? determines if the user should see a warning that results are not using natural language search.
+  # Shows when: user is opted-in AND viewing a Primo-only tab (which only supports keyword search).
+  def show_nls_warning?
+    @natural_language_search_optin && primo_tabs.include?(@active_tab)
   end
 end
--- a/app/views/search/results.html.erb
+++ b/app/views/search/results.html.erb
@@ -18,7 +18,6 @@
       <% elsif @results.present? && @errors.blank? %>
 
         <h2 class="results-context" data-matomo-seen="Search, Results Found, Tab: {{getActiveTabName}}"><%= results_summary(@pagination[:hits]) %></h2>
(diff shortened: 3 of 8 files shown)


app/controllers/search_controller.rb
```diff
@@ -16,7 +16,11 @@ def results
     # inject session preference for boolean type if it is present
     params[:booleanType] = cookies[:boolean_type] || 'AND'
 
+    # inject queryMode from cookie if not provided in URL params
```
JPrevost: Minor nit: this comment feels slightly misleading. We aren't injecting the queryMode from the cookie, we are injecting `hybrid` if the cookie evaluates to `true` but no url parameter is passed. It's minor difference but it made me have to read this several times to convince myself what it was doing was how I was interpreting it lol

app/controllers/search_controller.rb
```diff
@@ -442,4 +446,10 @@ def valid_token?
 
     false
   end
+
+  # show_nls_warning? determines if the user should see a warning that results are not using natural language search.
+  # Shows when: user is opted-in AND viewing a Primo-only tab (which only supports keyword search).
```
JPrevost, line 451: Not asking for a change now as I'm pretty sure this is what was asked for, but I'm curious if we should have a partial warning on the all tab like "Some results are not using natural language search" with a link to why. Let's leave this as-is for now though and we can bring this up to the project team.

app/controllers/application_controller.rb
```diff
@@ -1,6 +1,8 @@
 class ApplicationController < ActionController::Base
   helper Mitlibraries::Theme::Engine.helpers
 
+  before_action :set_natural_language_search_optin
```
JPrevost, line 4: There really a few aspects of this that I think are fine but we definitely need to confirm are innocuous in geo mode.

app/views/search/results.html.erb
```diff
@@ -18,7 +18,10 @@
       <% elsif @results.present? && @errors.blank? %>
 
         <h2 class="results-context" data-matomo-seen="Search, Results Found, Tab: {{getActiveTabName}}"><%= results_summary(@pagination[:hits]) %></h2>
-        <%= render partial: 'search/nls_alert' %>
+        <p class="results-context-description"><%= tab_description %></p>
```
JPrevost: Change requested:

I think this line got accidentally rebased into place and can be removed. It moved in a commit that just landed earlier today. This branch now shows the tab description twice, once in the new location near the tab as well as in this old location closer to the results.

test/controllers/natural_language_search_optin_test.rb
```diff
@@ -0,0 +1,310 @@
...
+    assert(cookies['natural_language_search_optin'].blank?)
+  end
+
+  test 'route redirects to return_to when provided' do
+    get '/natural_language_search_optin?natural_language_search_optin=true&return_to=/results?q=test'
+    assert_response :redirect
+    assert_redirected_to '/results?q=test'
+  end
+
+  test 'route redirects to root when no referer or return_to' do
```
JPrevost, line 82: Maybe a test for and illegal redirect to an external link goes to root?

test/controllers/natural_language_search_optin_test.rb
```diff
@@ -0,0 +1,310 @@
...
+    get '/natural_language_search_optin?natural_language_search_optin=false'
+
+    mock_primo_success
+    get '/results?q=test&tab=cdi'
+
+    assert_response :success
+    assert_select 'aside.nls-alert', count: 0
+  end
+
+  test 'warning does not show on articles tab with no opt-in cookie' do
```
JPrevost: Minor: these tests sometimes mention an articles tab and sometimes cdi. Consistency might make this easier to follow. I'd go with `cdi` if you do adjust as that should remain stable even if our labels change

JPrevost (changes requested):
The only required change is around a line accidentally reintroduced during a rebase.

The other comments are things you can tweak or not based on how they land with you.

Solid feature, I'm excited for others to get to try this soon!

app/views/search/results.html.erb
jazairi: Oops, good catch! Yes, that rebase was done hastily before review tagging.

app/controllers/search_controller.rb
jazairi: I agree, fwiw!

app/controllers/application_controller.rb
jazairi: I think GeoData is largely unaffected. It will set the cookie, but since GeoData doesn't use `queryMode`, it won't actually change the behavior. I also confirmed that nothing changes in the UI.

test/controllers/natural_language_search_optin_test.rb
```diff
@@ -85,6 +85,12 @@ def mock_primo_success
     assert_redirected_to root_path
   end
 
+  test 'external URLs redirect to route' do
```
JPrevost, line 88: Typo: I think you meant `root` here. Feel free to fixup without a re-review :)

JPrevost (approved):
:shipit:

PR #208, opened 2025-03-20
CloudFlare Turnstile

JPrevost:
Why are these changes being introduced:

* Bots are using more resources than desired

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/TIMX-480

How does this address that need:

* Adds rack-attack and bot_challenge_page gems
* Adds redis to heroku app.json (will need to manually enable in staging and production environments before merge)

##### Accessibility

- [ ] ANDI or WAVE has been run in accordance to [our guide](https://mitlibraries.github.io/guides/basics/a11y.html).
- [ ] This PR contains no changes to the view layer.
- [ ] New issues flagged by ANDI or WAVE have been resolved.
- [ ] New issues flagged by ANDI or WAVE have been ticketed (link in the Pull Request details above).
- [ ] No new accessibility issues have been flagged.

##### New ENV

- [ ] All new ENV is documented in README.
- [ ] All new ENV has been added to Heroku Pipeline, Staging and Prod.
- [ ] ENV has not changed.

##### Approval beyond code review

- [ ] UXWS/stakeholder approval has been confirmed.
- [ ] UXWS/stakeholder review will be completed retroactively.
- [ ] UXWS/stakeholder review is not needed.

##### Additional context needed to review

E.g., if the PR includes updated dependencies and/or data
migration, or how to confirm the feature is working.

##### Code

- [ ] I have confirmed that the code works as intended.
- [ ] Any CodeClimate issues have been fixed or confirmed as
added technical debt.

##### Documentation

- [ ] The commit message is clear and follows our guidelines
      (not just this pull request message).
- [ ] The documentation has been updated or is unnecessary.
- [ ] New dependencies are appropriate or there were no changes.

##### Testing

- [ ] There are appropriate tests covering any new functionality.
- [ ] No additional test coverage is required.

PR #436, opened 2026-07-23
Upgrade to Rails 8.1

jazairi:
##### Accessibility

##### New ENV

- [ ] ENV has not changed.

##### Code

added technical debt.

##### Documentation

##### Testing


config/initializers/content_security_policy.rb
```diff
@@ -4,22 +4,26 @@
...
+#   config.content_security_policy do |policy|
+#     policy.default_src :self, :https
+#     policy.font_src    :self, :https, :data
+#     policy.img_src     :self, :https, :data
+#     policy.object_src  :none
+#     policy.script_src  :self, :https
+#     policy.style_src   :self, :https
+#     # Specify URI for violation reports
+#     # policy.report_uri "/csp-violation-report-endpoint"
+#   end
```
JPrevost: Please review this more closely. We had explicit settings in the old config we want to carry forward while updating the boilerplate. CSP is confusing (to me) and I don't remember why we had some of the settings, but they were required for...something?

config/puma.rb
```diff
@@ -1,33 +1,42 @@
-# Thread per process count allows context switching on IO-bound tasks for better CPU utilization.
-threads_count = ENV.fetch('RAILS_MAX_THREADS') { 3 }
-threads(threads_count, threads_count)
-
-# Processes count, allows better CPU utilization when executing Ruby code.
-# Recommended to always run in at least one process so `rack-timeout` RACK_TERM_ON_TIMEOUT=1 can be used
-# https://devcenter.heroku.com/articles/h12-request-timeout-in-ruby-mri
-workers(ENV.fetch('WEB_CONCURRENCY') { 2 })
```
JPrevost, line 8: This one is confusing and I think comes from Heroku config. We set WEB_CONCURRENCY=1 in Heroku config and use rack-timeout (but don't seem to set RACK_TERM_ON_TIMEOUT=1 so 🤷🏻 ).

tldr; we may want to keep this around?

config/environments/production.rb
```diff
@@ -6,60 +6,37 @@
...
-  # config.assume_ssl = true
+  config.assume_ssl = true
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
   config.force_ssl = true
 
   # Skip http-to-https redirect for the default health check endpoint.
   # config.ssl_options = { redirect: { exclude: ->(request) { request.path == "/up" } } }
 
-  # Prepend all log lines with the following tags.
```
JPrevost, line 57: Hmmm. I thought we'd need to keep this to get the request id but it seems maybe it is now default or semantic logger adds it. Nothing to do here, just talking out loud :)

config/initializers/content_security_policy.rb
```diff
@@ -4,22 +4,26 @@
...
+#
+#   # Generate session nonces for permitted importmap, inline scripts, and inline styles.
+#   config.content_security_policy_nonce_generator = ->(request) { request.session.id.to_s }
+#   config.content_security_policy_nonce_directives = %w(script-src style-src)
+#
+#   # Automatically add `nonce` to `javascript_tag`, `javascript_include_tag`, and `stylesheet_link_tag`
+#   # if the corresponding directives are specified in `content_security_policy_nonce_directives`.
+#   # config.content_security_policy_nonce_auto = true
```
JPrevost, line 25: I wonder if this setting would actually be good (not as part of the upgrade). I think this may have been what the CSP above was trying to deal with now that I'm seeing this new option roll in. We were trying to set CSP (for some reason?) and had to include `unsafe-inline` to allow some JS to lad maybe... 🤷🏻 

It's probably worth a look through old Jira tickets and maybe opening new ones at some point in the future to try to better document what we are doing and why around CSP in which apps.

JPrevost (changes requested):
Overall looks good. I think we need to bring back our CSP and maybe reintroduce an option to Puma config.

Once we update those and land this, let's check GeoData in staging closely and then tag #proj_use to do some testing in unified search staging before we promote.

config/environments/production.rb
jazairi: Yup, semantic logger does this now.

config/puma.rb
jazairi: I'm not sure I agree. I think Heroku config should come from Heroku, not hardcoded in Puma config. Furthermore, a fallback of `2` feels oddly expensive to me. I think that if we want a setting of `2` in staging and prod, we should set that explicitly in our staging/prod apps.

To be fair, this is another instance of "I'm not sure why we did this," so the safest option is to keep it as-is. But, imo, the correct option is likely to modify our staging/prod apps to use the config we want.

config/initializers/content_security_policy.rb
jazairi: Yeah, you're right. It was implemented [here](https://github.com/MITLibraries/timdex-ui/commit/c69842e) and enabled [here](https://github.com/MITLibraries/timdex-ui/commit/73d9b3a). I'm not sure why it's needed either, but I agree a) that we should keep it in for now, and b) we should explore nonce-based CSPs.

config/puma.rb
JPrevost: Setting this in heroku config cannot affect anything unless we tell puma to read the value though. Heroku is telling us how to best run Puma in Heroku.

These tickets clarify somewhat why that setting is in place:
- https://mitlibraries.atlassian.net/browse/USE-396
- https://mitlibraries.atlassian.net/browse/TIMX-298 (yes, this is a ticket that no work was logged under but I'm pretty sure I just forgot there was a ticket when I did the work and noticed that recently so I closed it)

JPrevost: Also, the 2 value isn't how many dynos are running it's how many instances of the app run in each dyno. This allows us to most effectively use the RAM allocated in each dyno. Our dynos are so small we don't get many instances.

config/initializers/content_security_policy.rb
jazairi: There isn't much info in [the PR](https://github.com/MITLibraries/timdex-ui/pull/208) or [the ticket](https://mitlibraries.atlassian.net/browse/TIMX-480). In retrospect, I wish I'd asked about the CSP rather than accepting it at face value before approving. The "what" is easy enough to figure out, but the "why" is probably lost to time.

I opened a ticket to explore [implementing nonce-based CSP](https://mitlibraries.atlassian.net/browse/USE-661), and might start working on it once this merges.

config/puma.rb
jazairi: Ahh, gotcha. Thanks for clarifying.

config/puma.rb
```diff
@@ -1,33 +1,47 @@
...
+# Any libraries that use a connection pool or another resource pool should
+# be configured to provide at least as many connections as the number of
+# threads. This includes Active Record's `pool` parameter in `database.yml`.
+threads_count = ENV.fetch("RAILS_MAX_THREADS", 3)
+threads threads_count, threads_count
 
 # Processes count, allows better CPU utilization when executing Ruby code.
 # Recommended to always run in at least one process so `rack-timeout` RACK_TERM_ON_TIMEOUT=1 can be used
 # https://devcenter.heroku.com/articles/h12-request-timeout-in-ruby-mri
 workers(ENV.fetch('WEB_CONCURRENCY') { 2 })
```
JPrevost, line 34: Based on our comments about this, I'd be pretty good with defaulting to 1 and then instances that want more can intentionally bump it up. What do you think about that?

JPrevost (approved):
Approving and also totally open to change the fall back for WEB_CONCURRENCY to 1 based on what you said about it along the way.

config/puma.rb
jazairi: If 2 is what is broadly needed, I trust that it's a reasonable fallback. It doesn't seem like it's causing issues thus far.

